PR #222, opened 2023-01-22
Add `DelegateCall` syscall.

elintul:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/222)

### DIFF
--- a/crates/blockifier/src/execution/entry_point.rs
+++ b/crates/blockifier/src/execution/entry_point.rs
@@ -11,6 +11,7 @@ use crate::block_context::BlockContext;
 use crate::execution::contract_class::ContractClass;
 use crate::execution::errors::{EntryPointExecutionError, PreExecutionError};
 use crate::execution::execution_utils::execute_entry_point_call;
+use crate::execution::syscall_handling::SyscallHintProcessor;
 use crate::state::state_api::State;
 use crate::transaction::objects::AccountTransactionContext;
 
@@ -61,6 +62,27 @@ impl CallEntryPoint {
         match entry_points_of_same_type.iter().find(|ep| ep.selector == self.entry_point_selector) {
             Some(entry_point) => Ok(entry_point.offset.0),
             None => Err(PreExecutionError::EntryPointNotFound(self.entry_point_selector)),
+        }
+    }
+
+    pub fn create_for_library_call(
+        class_hash: ClassHash,
+        call_to_external: bool,
+        entry_point_selector: EntryPointSelector,
+        calldata: Calldata,
+        syscall_handler: &SyscallHintProcessor<'_>,
+    ) -> Self {
+        let entry_point_type =
+            if call_to_external { EntryPointType::External } else { EntryPointType::L1Handler };
+
+        CallEntryPoint {
+            class_hash: Some(class_hash),
+            entry_point_type,
+            entry_point_selector,
+            calldata,
+            // The call context remains the same in a library call.
+            storage_address: syscall_handler.storage_address,
+            caller_address: syscall_handler.caller_address,
         }
     }
 }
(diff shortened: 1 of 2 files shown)


elintul (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/222)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @alonh5 and @noaov1)

___
*[`crates/blockifier/src/execution/entry_point.rs` line 68 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/222#-NMSoNyx0Jb4tjpqN1F0:-NMTDwoP5pG6uSKSSCYN:b9v3aoh) ([raw file](https://github.com/starkware-libs/blockifier/blob/247795b672d1beb0470f6da967e0e963888b5ad8/crates/blockifier/src/execution/entry_point.rs#L68)):*
<details><summary><i>Previously, alonh5 (Alon Haramati) wrote…</i></summary><blockquote>

Maybe move to `syscall_handling.rs`? (You can create another impl block)
</blockquote></details>

You think it's better to split `impl` blocks of the same struct?
Talked to OriZ, he says that pure `impl`s of a struct (not traits) should appear next to the struct, and if they're not in the same crate - it won't compile. 
This is a type of a c-tor for `CallEntryPoint`, and we're also going to use it for tests, so I think we should leave it here.
___
*[`crates/blockifier/src/execution/syscalls.rs` line 246 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/222#-NMScxi81v8Mdj0iAmi9:-NMTHB_H3MpkBk15djQg:b-vqjxfe) ([raw file](https://github.com/starkware-libs/blockifier/blob/247795b672d1beb0470f6da967e0e963888b5ad8/crates/blockifier/src/execution/syscalls.rs#L246)):*
<details><summary><i>Previously, alonh5 (Alon Haramati) wrote…</i></summary><blockquote>

Shouldn't this be:
(Same for LibraryCall)
</blockquote></details>

Aligned aliases throughout the file; not sure it's worth it, but alignment was due (either all have, or all don't).

alonh5 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/222#-:-NMTJRtyAN0qL8XIm-Zb:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @elintul and @noaov1)

___
*[`crates/blockifier/src/execution/entry_point.rs` line 68 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/222#-NMSoNyx0Jb4tjpqN1F0:-NMTHoQM0TYFGIup56JL:bibymxb) ([raw file](https://github.com/starkware-libs/blockifier/blob/247795b672d1beb0470f6da967e0e963888b5ad8/crates/blockifier/src/execution/entry_point.rs#L68)):*

You think it's better to split `impl` blocks of the same struct?
Talked to OriZ, he says that pure `impl`s of a struct (not traits) should appear next to the struct, and if they're not in the same crate - it won't compile. 
This is a type of a c-tor for `CallEntryPoint`, and we're also going to use it for tests, so I think we should leave it here.
</blockquote></details>

I think there's a case for both. A good example is `create_for testing`, which should be in the testing file IMO. Here, seeing the syscall handler import addition, I thought it might be right to move it there because it should only be used there. Your call :)
___
*[`crates/blockifier/src/execution/syscalls.rs` line 246 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/222#-NMScxi81v8Mdj0iAmi9:-NMTIjs_611Jq3GyGtaj:b-ufm0c9) ([raw file](https://github.com/starkware-libs/blockifier/blob/247795b672d1beb0470f6da967e0e963888b5ad8/crates/blockifier/src/execution/syscalls.rs#L246)):*

Aligned aliases throughout the file; not sure it's worth it, but alignment was due (either all have, or all don't).
</blockquote></details>

Great :)

PR #363, opened 2023-03-28
Add get_sorted_l2_to_l1_payloads_length.

noaov1:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/363)

### DIFF
--- a/crates/blockifier/src/execution/entry_point.rs
+++ b/crates/blockifier/src/execution/entry_point.rs
@@ -2,7 +2,7 @@ use std::collections::HashSet;
 
 use cairo_vm::vm::runners::cairo_runner::ExecutionResources as VmExecutionResources;
 use starknet_api::core::{ClassHash, ContractAddress, EntryPointSelector};
-use starknet_api::hash::StarkFelt;
+use starknet_api::hash::{StarkFelt, StarkHash};
 use starknet_api::state::{EntryPoint, EntryPointType, StorageKey};
 use starknet_api::transaction::{Calldata, EthAddress, EventContent, L2ToL1Payload};
 
@@ -14,6 +14,7 @@ use crate::execution::errors::{EntryPointExecutionError, PreExecutionError};
 use crate::execution::execution_utils::execute_entry_point_call;
 use crate::execution::syscall_handling::SyscallCounter;
 use crate::state::state_api::State;
+use crate::transaction::errors::TransactionExecutionError;
 use crate::transaction::objects::AccountTransactionContext;
 
 #[cfg(test)]
@@ -106,7 +107,7 @@ impl CallEntryPoint {
             match entry_points_of_same_type.get(0) {
                 Some(entry_point) => {
                     if entry_point.selector
-                        == EntryPointSelector(DEFAULT_ENTRY_POINT_SELECTOR.into())
+                        == EntryPointSelector(StarkHash::from(DEFAULT_ENTRY_POINT_SELECTOR))
                     {
                         return Ok(entry_point.offset.0);
                     } else {
@@ -178,6 +179,32 @@ pub struct CallInfo {
     // Additional information gathered during execution.
     pub storage_read_values: Vec<StarkFelt>,
     pub accessed_storage_keys: HashSet<StorageKey>,
+}
+
+impl CallInfo {
+    /// Returns a list of StarkNet L2ToL1Payload length collected during the execution, sorted
+    /// by the order in which they were sent.
+    pub fn get_sorted_l2_to_l1_payloads_length(
+        &self,
+    ) -> Result<Vec<usize>, TransactionExecutionError> {
+        let n_messages = self.into_iter().map(|call| call.execution.l2_to_l1_messages.len()).sum();
+        let mut starknet_l2_to_l1_payloads_length: Vec<Option<usize>> = vec![None; n_messages];
+
+        for call in self.into_iter() {
+            for ordered_message_content in &call.execution.l2_to_l1_messages {
+                if starknet_l2_to_l1_payloads_length[ordered_message_content.order].is_some() {
+                    return Err(TransactionExecutionError::UnexpectedHoles {
+                        object: "L2-to-L1 message".to_string(),
+                        order: ordered_message_content.order,
+                    });
+                }
+                starknet_l2_to_l1_payloads_length[ordered_message_content.order] =
+                    Some(ordered_message_content.message.payload.0.len());
+            }
+        }
+
+        Ok(starknet_l2_to_l1_payloads_length.into_iter().flatten().collect())
+    }
 }
 
 pub struct CallInfoIter<'a> {
--- a/crates/blockifier/src/transaction/errors.rs
+++ b/crates/blockifier/src/transaction/errors.rs
@@ -47,6 +47,8 @@ pub enum TransactionExecutionError {
     StateError(#[from] StateError),
     #[error("Calling other contracts during '{entry_point_kind}' execution is forbidden.")]
     UnauthorizedInnerCall { entry_point_kind: String },
+    #[error("Unexpected holes in the {object} order. Two objects with the same order: {order}.")]
+    UnexpectedHoles { object: String, order: usize },
     #[error("Unknown chain ID '{chain_id:?}'.")]
     UnknownChainId { chain_id: String },
 }


noaov1 (commented):
@amosStarkware  FYI

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/363#-:-NRbXt5z1EEvoJEa4J6-:bpq2vt7)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @dorimedini-starkware and @giladchase)

___
*[`crates/blockifier/src/execution/entry_point.rs` line 188 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/363#-NRbXTrbBJnkiMaysIEx:-NRbXTrbBJnkiMaysIEy:b-85b7k0) ([raw file](https://github.com/starkware-libs/blockifier/blob/1af45a3f3631b07f1c0402e4f8d7ced01876112b/crates/blockifier/src/execution/entry_point.rs#L188)):*
> ```XML
>     /// Returns a list of StarkNet L2ToL1Payload length collected during the execution, sorted
>     /// by the order in which they were sent.
>     fn _get_sorted_l2_to_l1_payloads_length(
> ```

For the fee flow calculation it does not have to be sorted, but it is, so I left the name.



_Code quote:_
```XML
_get_sorted_l2_to_l1_payloads_length
```
___
*[`crates/blockifier/src/transaction/errors.rs` line 51 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/363#-NRbXdJwA0HK0hXDvQ9w:-NRbXdJwA0HK0hXDvQ9x:b-a8bs0a) ([raw file](https://github.com/starkware-libs/blockifier/blob/1af45a3f3631b07f1c0402e4f8d7ced01876112b/crates/blockifier/src/transaction/errors.rs#L51)):*
> ```XML
>     UnauthorizedInnerCall { entry_point_kind: String },
>     #[error("Unexpected holes in the {object} order.")]
>     UnexpectedHoles { object: String },
> ```

Not sure if this is the right place. I'll see later and we will use `_get_sorted_l2_to_l1_payloads_length` and change it if needed.



_Code quote:_
```XML
    #[error("Unexpected holes in the {object} order.")]
    UnexpectedHoles { object: String },
```

dorimedini-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/363)* status: all files reviewed, 3 unresolved discussions (waiting on @giladchase and @noaov1)

___
*[`crates/blockifier/src/execution/entry_point.rs` line 188 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/363#-NRbXTrbBJnkiMaysIEx:-NRbZFEI-IbqjykuWOlq:b-5waera) ([raw file](https://github.com/starkware-libs/blockifier/blob/1af45a3f3631b07f1c0402e4f8d7ced01876112b/crates/blockifier/src/execution/entry_point.rs#L188)):*

For the fee flow calculation it does not have to be sorted, but it is, so I left the name.
</blockquote></details>

why the leading underscore? it's not marked `pub`, that should be enough in rust, no?
___
*[`crates/blockifier/src/execution/entry_point.rs` line 194 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/363#-NRb_Pc35jR1SjjFabUv:-NRb_Pc35jR1SjjFabUw:bo9568) ([raw file](https://github.com/starkware-libs/blockifier/blob/1af45a3f3631b07f1c0402e4f8d7ced01876112b/crates/blockifier/src/execution/entry_point.rs#L194)):*
> ```XML
>         let mut starknet_l2_to_l1_messages: Vec<Option<usize>> = vec![None; n_messages];
> 
>         for call in self.into_iter() {
> ```

I'd like to understand what the assumptions are in the message orders...
1. If we assume nothing, then shouldn't add a check that each order is less than n_messages?
2. If we assume all orders are less than n_messages, then there is a hole if and only if two messages have the same order, right? In that case, wouldn't it be more informative to catch this when we try to assign a message to `starknet_l2_to_l1_messages` at an index without `None`? That way we can return an error with the actual problematic message (or at least it's order)

WDYT?



_Code quote:_
```XML
        let n_messages = self.into_iter().map(|call| call.execution.l2_to_l1_messages.len()).sum();
        let mut starknet_l2_to_l1_messages: Vec<Option<usize>> = vec![None; n_messages];

        for call in self.into_iter() {
            for ordered_message_content in &call.execution.l2_to_l1_messages {
                starknet_l2_to_l1_messages[ordered_message_content.order] =
                    Some(ordered_message_content.message.payload.0.len());
            }
        }
        if starknet_l2_to_l1_messages.iter().contains(&None) {
            return Err(TransactionExecutionError::UnexpectedHoles {
                object: "L2-to-L1 message".to_string(),
            });
        }

        Ok(starknet_l2_to_l1_messages.into_iter().flatten().collect())
```

noaov1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/363)* status: all files reviewed, 3 unresolved discussions (waiting on @dorimedini-starkware and @giladchase)

___
*[`crates/blockifier/src/execution/entry_point.rs` line 188 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/363#-NRbXTrbBJnkiMaysIEx:-NRcCaoO48ksex7GyQzn:bvxyqti) ([raw file](https://github.com/starkware-libs/blockifier/blob/1af45a3f3631b07f1c0402e4f8d7ced01876112b/crates/blockifier/src/execution/entry_point.rs#L188)):*

why the leading underscore? it's not marked `pub`, that should be enough in rust, no?
</blockquote></details>

On the contrary. It is because I didn't have `pub` and I didn't use it. Will add `pub`.
___
*[`crates/blockifier/src/execution/entry_point.rs` line 194 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/363#-NRb_Pc35jR1SjjFabUv:-NRcJo6i6ccQh4PGGY6t:b3uliq7) ([raw file](https://github.com/starkware-libs/blockifier/blob/1af45a3f3631b07f1c0402e4f8d7ced01876112b/crates/blockifier/src/execution/entry_point.rs#L194)):*

I'd like to understand what the assumptions are in the message orders...
1. If we assume nothing, then shouldn't add a check that each order is less than n_messages?
2. If we assume all orders are less than n_messages, then there is a hole if and only if two messages have the same order, right? In that case, wouldn't it be more informative to catch this when we try to assign a message to `starknet_l2_to_l1_messages` at an index without `None`? That way we can return an error with the actual problematic message (or at least it's order)

WDYT?
</blockquote></details>

1. What do you mean by "assuming nothing"? And wouldn't it fail otherwise? (in `starknet_l2_to_l1_messages[ordered_message_content.order] = Some(ordered_message_content.message.payload.0.len());`
2. Do you think that the specific order of the collision will be informative? I don't mind. But, in that case you repeat this check `n_messages` times (instead of one time).

dorimedini-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/363)* status: all files reviewed, 3 unresolved discussions (waiting on @giladchase and @noaov1)

___
*[`crates/blockifier/src/execution/entry_point.rs` line 194 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/363#-NRb_Pc35jR1SjjFabUv:-NRcOqRa05bDr1EgpBxr:b-9wu3ld) ([raw file](https://github.com/starkware-libs/blockifier/blob/1af45a3f3631b07f1c0402e4f8d7ced01876112b/crates/blockifier/src/execution/entry_point.rs#L194)):*

1. What do you mean by "assuming nothing"? And wouldn't it fail otherwise? (in `starknet_l2_to_l1_messages[ordered_message_content.order] = Some(ordered_message_content.message.payload.0.len());`
2. Do you think that the specific order of the collision will be informative? I don't mind. But, in that case you repeat this check `n_messages` times (instead of one time).
</blockquote></details>

1. This will fail with panic, right? I'm saying, if we don't assume the message orders are correct, why not return an `Err` if the index is out of bounds?
2. Do you think it will be slower than `starknet_l2_to_l1_messages.iter().contains(&None)`? This is also linear IIUC. It's more informative, especially since you know which message collides (you can output the hash as well)

noaov1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/363)* status: all files reviewed, 3 unresolved discussions (waiting on @dorimedini-starkware and @giladchase)

___
*[`crates/blockifier/src/execution/entry_point.rs` line 194 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/363#-NRb_Pc35jR1SjjFabUv:-NRgWqlg5sfwcqUCeLcC:b6jzrij) ([raw file](https://github.com/starkware-libs/blockifier/blob/1af45a3f3631b07f1c0402e4f8d7ced01876112b/crates/blockifier/src/execution/entry_point.rs#L194)):*

1. This will fail with panic, right? I'm saying, if we don't assume the message orders are correct, why not return an `Err` if the index is out of bounds?
2. Do you think it will be slower than `starknet_l2_to_l1_messages.iter().contains(&None)`? This is also linear IIUC. It's more informative, especially since you know which message collides (you can output the hash as well)
</blockquote></details>

1. It will indeed fail with panic, but I don't think this can happen ( `index > n_massages`). We added this check to see that no two messages received the same order number.
2. No. I wouldn't know the hash of the two messages (I save the payload length. I can print the new message). WDYT?

dorimedini-starkware (approved):
:lgtm:

> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/363#-:-NRhE2A92btp6Hb11wYN:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @alonh5 and @giladchase)

___
*[`crates/blockifier/src/execution/entry_point.rs` line 194 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/363#-NRb_Pc35jR1SjjFabUv:-NRhDbE_7hOw9Jh6P0F-:bm19cog) ([raw file](https://github.com/starkware-libs/blockifier/blob/1af45a3f3631b07f1c0402e4f8d7ced01876112b/crates/blockifier/src/execution/entry_point.rs#L194)):*

1. It will indeed fail with panic, but I don't think this can happen ( `index > n_massages`). We added this check to see that no two messages received the same order number.
2. No. I wouldn't know the hash of the two messages (I save the payload length. I can print the new message). WDYT?
</blockquote></details>

I like it like this :)

PR #386, opened 2023-04-04
Add `calculate_tx_resources`.

noaov1:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/386)

### DIFF
--- a/crates/blockifier/src/execution/entry_point.rs
+++ b/crates/blockifier/src/execution/entry_point.rs
@@ -16,7 +16,7 @@ use crate::execution::execution_utils::execute_entry_point_call;
 use crate::execution::syscall_handling::SyscallCounter;
 use crate::state::state_api::State;
 use crate::transaction::errors::TransactionExecutionError;
-use crate::transaction::objects::AccountTransactionContext;
+use crate::transaction::objects::{AccountTransactionContext, TransactionExecutionResult};
 
 #[cfg(test)]
 #[path = "entry_point_test.rs"]
@@ -185,9 +185,7 @@ pub struct CallInfo {
 impl CallInfo {
     /// Returns a list of StarkNet L2ToL1Payload length collected during the execution, sorted
     /// by the order in which they were sent.
-    pub fn get_sorted_l2_to_l1_payloads_length(
-        &self,
-    ) -> Result<Vec<usize>, TransactionExecutionError> {
+    pub fn get_sorted_l2_to_l1_payloads_length(&self) -> TransactionExecutionResult<Vec<usize>> {
         let n_messages = self.into_iter().map(|call| call.execution.l2_to_l1_messages.len()).sum();
         let mut starknet_l2_to_l1_payloads_length: Vec<Option<usize>> = vec![None; n_messages];
 
--- a/crates/blockifier/src/fee/fee_utils.rs
+++ b/crates/blockifier/src/fee/fee_utils.rs
@@ -3,8 +3,61 @@ use std::collections::{HashMap, HashSet};
 use itertools::max;
 
 use crate::block_context::BlockContext;
+use crate::execution::entry_point::{CallInfo, ExecutionResources};
+use crate::fee::gas_usage::calculate_tx_gas_usage;
+use crate::fee::os_usage::get_additional_os_resources;
+use crate::state::cached_state::TransactionalState;
+use crate::state::state_api::StateReader;
 use crate::transaction::errors::TransactionExecutionError;
 use crate::transaction::objects::{ResourcesMapping, TransactionExecutionResult};
+use crate::transaction::transaction_types::TransactionType;
+
+const FEE_TRANSFER_N_STORAGE_CHANGES: u8 = 2; // Sender and sequencer balance update.
+// Exclude the sequencer balance update, since it's charged once throughout the batch.
+const FEE_TRANSFER_N_STORAGE_CHANGES_TO_CHARGE: u8 = FEE_TRANSFER_N_STORAGE_CHANGES - 1;
+
+/// Returns the total resources needed to include the most recent transaction in a StarkNet batch
+/// (recent w.r.t. application on the given state) i.e., L1 gas usage and Cairo execution resources.
+/// Used for transaction fee; calculation is made as if the transaction is the first in batch, for
+/// consistency.
+pub fn calculate_tx_resources<S: StateReader>(
+    resources_manager: ExecutionResources,
+    call_infos: &[Option<&CallInfo>],
+    tx_type: TransactionType,
+    state: &mut TransactionalState<'_, S>,
+    l1_handler_payload_size: Option<usize>,
+) -> TransactionExecutionResult<ResourcesMapping> {
+    let (n_storage_changes, n_modified_contracts, n_class_updates) =
+        state.count_actual_state_changes();
+
+    let non_optional_call_infos: Vec<&CallInfo> =
+        call_infos.iter().flat_map(|&optional_call_info| optional_call_info).collect();
+
+    let mut l2_to_l1_payloads_length = vec![];
+    for call_info in non_optional_call_infos {
+        l2_to_l1_payloads_length.extend(call_info.get_sorted_l2_to_l1_payloads_length()?);
+    }
+
+    let l1_gas_usage = calculate_tx_gas_usage(
+        &l2_to_l1_payloads_length,
+        n_modified_contracts,
+        n_storage_changes + usize::from(FEE_TRANSFER_N_STORAGE_CHANGES_TO_CHARGE),
+        l1_handler_payload_size,
+        n_class_updates,
+    );
+
+    let mut cairo_usage = resources_manager.vm_resources;
+    // Add additional Cairo resources needed for the OS to run the transaction.
+    cairo_usage += &get_additional_os_resources(resources_manager.syscall_counter, tx_type)?;
+    cairo_usage = cairo_usage.filter_unused_builtins();
+    let mut tx_resources = HashMap::from([
+        ("l1_gas_usage".to_string(), l1_gas_usage),
+        ("n_steps".to_string(), cairo_usage.n_steps + cairo_usage.n_memory_holes),
+    ]);
+    tx_resources.extend(cairo_usage.builtin_instance_counter);
+
+    Ok(ResourcesMapping(tx_resources))
+}
 
 pub fn extract_l1_gas_and_cairo_usage(
     resources: &ResourcesMapping,


dorimedini-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/386)* status: all files reviewed, 3 unresolved discussions (waiting on @alonh5 and @noaov1)

___
*[`crates/blockifier/src/fee/fee_utils.rs` line 13 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/386#-NSBIcpi3h2tc-I371vY:-NSBIcpi3h2tc-I371vZ:b-250ayg) ([raw file](https://github.com/starkware-libs/blockifier/blob/999210ab84a841be4068142dac94870f3e0b6e76/crates/blockifier/src/fee/fee_utils.rs#L13)):*
> ```XML
> const FEE_TRANSFER_N_STORAGE_CHANGES: u32 = 2; // Sender and sequencer balance update.
> // Exclude the sequencer balance update, since it's charged once throughout the batch.
> const FEE_TRANSFER_N_STORAGE_CHANGES_TO_CHARGE: u32 = FEE_TRANSFER_N_STORAGE_CHANGES - 1;
> ```

maybe we should have only two integer types... `usize` or `u128`.
I suggest that for any integer variable `X`:
1. If there exists some data structure with `~X` elements (`Vec`, `HashMap` etc), then `X` should be of type `usize`.
2. Otherwise, `X` should be `u128`.
WDYT?
Specifically, these two seem like they should be `usize`, because we *will* have some form of list/map of all storage changes, right?



_Code quote:_
```XML
const FEE_TRANSFER_N_STORAGE_CHANGES: u32 = 2; // Sender and sequencer balance update.
// Exclude the sequencer balance update, since it's charged once throughout the batch.
const FEE_TRANSFER_N_STORAGE_CHANGES_TO_CHARGE: u32 = FEE_TRANSFER_N_STORAGE_CHANGES - 1;
```
___
*[`crates/blockifier/src/fee/fee_utils.rs` line 29 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/386#-NSBIKY0-Nx1aZ_PlmGc:-NSBIKY0-Nx1aZ_PlmGd:bf56h37) ([raw file](https://github.com/starkware-libs/blockifier/blob/999210ab84a841be4068142dac94870f3e0b6e76/crates/blockifier/src/fee/fee_utils.rs#L29)):*
> ```XML
>         state.count_actual_state_changes();
> 
>     let non_optional_call_infos: Vec<&CallInfo> = call_infos.iter().flatten().collect();
> ```

what's this? on python side it seems you should be filtering `None`s, not flattening



_Code quote:_
```XML
call_infos.iter().flatten().collect()
```
___
*[`crates/blockifier/src/fee/fee_utils.rs` line 49 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/386#-NSBJVKJCUZhOBjcQZPK:-NSBJVKJCUZhOBjcQZPL:b2l7ffp) ([raw file](https://github.com/starkware-libs/blockifier/blob/999210ab84a841be4068142dac94870f3e0b6e76/crates/blockifier/src/fee/fee_utils.rs#L49)):*
> ```XML
>     let mut tx_resources = HashMap::from([
>         ("l1_gas_usage".to_string(), l1_gas_usage),
>         ("n_steps".to_string(), cairo_usage.n_steps + cairo_usage.n_memory_holes),
> ```

Perhaps these constant strings should be named constants...? WDYT?
`"l1_gas_usage"` is used in this file already, and `"n_steps"` is used to define the JSON in `os_resources.rs`



_Code quote:_
```XML
        ("l1_gas_usage".to_string(), l1_gas_usage),
        ("n_steps".to_string(), cairo_usage.n_steps + cairo_usage.n_memory_holes),
```

dorimedini-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/386)* status: all files reviewed, 4 unresolved discussions (waiting on @alonh5 and @noaov1)

___
*[`crates/blockifier/src/fee/fee_utils.rs` line 49 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/386#-NSBLIif4t_KFbwkcRLA:-NSBLIif4t_KFbwkcRLB:b-nuu8ss) ([raw file](https://github.com/starkware-libs/blockifier/blob/999210ab84a841be4068142dac94870f3e0b6e76/crates/blockifier/src/fee/fee_utils.rs#L49)):*
> ```XML
>     let mut tx_resources = HashMap::from([
>         ("l1_gas_usage".to_string(), l1_gas_usage),
>         ("n_steps".to_string(), cairo_usage.n_steps + cairo_usage.n_memory_holes),
> ```

this is part of `to_dict` on python, maybe add a comment here why memory holes are added as `n_steps` here?



_Code quote:_
```XML
cairo_usage.n_steps + cairo_usage.n_memory_holes
```

noaov1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/386)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @alonh5 and @dorimedini-starkware)

___
*[`crates/blockifier/src/fee/fee_utils.rs` line 13 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/386#-NSBIcpi3h2tc-I371vY:-NSWGGZGAJeJ_cmHxAn-:bnzjs6t) ([raw file](https://github.com/starkware-libs/blockifier/blob/999210ab84a841be4068142dac94870f3e0b6e76/crates/blockifier/src/fee/fee_utils.rs#L13)):*

maybe we should have only two integer types... `usize` or `u128`.
I suggest that for any integer variable `X`:
1. If there exists some data structure with `~X` elements (`Vec`, `HashMap` etc), then `X` should be of type `usize`.
2. Otherwise, `X` should be `u128`.
WDYT?
Specifically, these two seem like they should be `usize`, because we *will* have some form of list/map of all storage changes, right?
</blockquote></details>

I see what you mean. Though I read [this] (https://www.reddit.com/r/rust/comments/4lql5l/beginner_question_should_i_use_usize_and_isize/) and it seems that we should use a concrete type and maybe convert the `usize` using `try_from()`. WDYT? 
___
*[`crates/blockifier/src/fee/fee_utils.rs` line 29 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/386#-NSBIKY0-Nx1aZ_PlmGc:-NSWGnqo1F_ztWmuy9mV:bu7ivke) ([raw file](https://github.com/starkware-libs/blockifier/blob/999210ab84a841be4068142dac94870f3e0b6e76/crates/blockifier/src/fee/fee_utils.rs#L29)):*

what's this? on python side it seems you should be filtering `None`s, not flattening
</blockquote></details>

Flattening also filters `None`s. Since `Option` implements `IntoIterator`, we can convert our `Vec` into an iterator (which will be an iterator of iterators) and then flatten it.
___
*[`crates/blockifier/src/fee/fee_utils.rs` line 49 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/386#-NSBJVKJCUZhOBjcQZPK:-NSWKcKm3vJiVP_9zD9K:bqlnojg) ([raw file](https://github.com/starkware-libs/blockifier/blob/999210ab84a841be4068142dac94870f3e0b6e76/crates/blockifier/src/fee/fee_utils.rs#L49)):*

Perhaps these constant strings should be named constants...? WDYT?
`"l1_gas_usage"` is used in this file already, and `"n_steps"` is used to define the JSON in `os_resources.rs`
</blockquote></details>

Do you think it is necessary? Wouldn't that be more confusing?

dorimedini-starkware (commented):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/386)* status: all files reviewed, 3 unresolved discussions (waiting on @alonh5 and @noaov1)

___
*[`crates/blockifier/src/fee/fee_utils.rs` line 13 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/386#-NSBIcpi3h2tc-I371vY:-NSeIhqH5ZmXhNWWSUMY:b-rscwvd) ([raw file](https://github.com/starkware-libs/blockifier/blob/999210ab84a841be4068142dac94870f3e0b6e76/crates/blockifier/src/fee/fee_utils.rs#L13)):*

I see what you mean. Though I read [this] (https://www.reddit.com/r/rust/comments/4lql5l/beginner_question_should_i_use_usize_and_isize/) and it seems that we should use a concrete type and maybe convert the `usize` using `try_from()`. WDYT? 
</blockquote></details>

in the blockifier we hit an error on block 256 due to papyrus not working on a different architecture, so maybe you're right.
but why are you so sure there are at most 256 storage changes for fee transfer?
___
*[`crates/blockifier/src/fee/fee_utils.rs` line 49 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/386#-NSBJVKJCUZhOBjcQZPK:-NSeFosi-UJsFtdntM0H:b4os34y) ([raw file](https://github.com/starkware-libs/blockifier/blob/999210ab84a841be4068142dac94870f3e0b6e76/crates/blockifier/src/fee/fee_utils.rs#L49)):*

Do you think it is necessary? Wouldn't that be more confusing?
</blockquote></details>

shouldn't be, if the constant has a good name (N_STEPS_KEY_STRING or something).
makes it less error-prone.
non-blocking anyway, your call

noaov1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/386)* status: all files reviewed, 3 unresolved discussions (waiting on @alonh5, @dorimedini-starkware, and @noaov1)

___
*[`crates/blockifier/src/fee/fee_utils.rs` line 13 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/386#-NSBIcpi3h2tc-I371vY:-NSeObhT7S4a8CwrfQd_:b-yo0msu) ([raw file](https://github.com/starkware-libs/blockifier/blob/999210ab84a841be4068142dac94870f3e0b6e76/crates/blockifier/src/fee/fee_utils.rs#L13)):*

in the blockifier we hit an error on block 256 due to papyrus not working on a different architecture, so maybe you're right.
but why are you so sure there are at most 256 storage changes for fee transfer?
</blockquote></details>

I decided to change every numeric value to `u64` (after consulting with Elin that the numbers should not exceed `2^64` bits) (except for the gas usage that is `u128`). I chose here (as a temporary solution) `u8` because the size of a `usize` is always guaranteed to be 8 bits or greater and a `u8` to `usize` From conversion will always exist. 
___
*[`crates/blockifier/src/fee/fee_utils.rs` line 49 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/386#-NSBJVKJCUZhOBjcQZPK:-NSeQILL720_b_CVIoyJ:b-bsljaj) ([raw file](https://github.com/starkware-libs/blockifier/blob/999210ab84a841be4068142dac94870f3e0b6e76/crates/blockifier/src/fee/fee_utils.rs#L49)):*

shouldn't be, if the constant has a good name (N_STEPS_KEY_STRING or something).
makes it less error-prone.
non-blocking anyway, your call
</blockquote></details>

Got it. I agree. Will do it in my next PR.
___
*[`crates/blockifier/src/fee/fee_utils.rs` line 49 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/386#-NSBLIif4t_KFbwkcRLA:-NSeQOzF52vu5iFU1zro:b-f0h5ug) ([raw file](https://github.com/starkware-libs/blockifier/blob/999210ab84a841be4068142dac94870f3e0b6e76/crates/blockifier/src/fee/fee_utils.rs#L49)):*

this is part of `to_dict` on python, maybe add a comment here why memory holes are added as `n_steps` here?
</blockquote></details>

Will do. Still waiting for an answer as to why this is.

dorimedini-starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/386)* status: all files reviewed, 1 unresolved discussion (waiting on @alonh5 and @noaov1)

PR #427, opened 2023-04-24
feature: replace ctor by lazy_static

tdelabro:
ctor is not supported for madara wasm target
So let's use lazy static instead

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/427)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -244,10 +244,10 @@ dependencies = [
  "assert_matches",
  "cairo-felt",
  "cairo-vm",
- "ctor 0.2.0",
  "derive_more",
  "indexmap",
  "itertools",
+ "lazy_static",
  "log",
  "num-bigint",
  "num-integer",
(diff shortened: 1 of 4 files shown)


Oppen:
I'd suggest using [once_cell](https://crates.io/crates/once_cell) instead, as it will be incorporated into the stdlib soon (meaning one fewer dep).

tdelabro:
Once cell is not compatible with no_std for now: https://docs.rs/once_cell/1.17.1/once_cell/#faq (second line)
So I would prefer you avoid it, as I will have to change it in my fork

Oppen:
> Once cell is not compatible with no_std for now: https://docs.rs/once_cell/1.17.1/once_cell/#faq (second line)
So I would prefer you avoid it, as I will have to change it in my fork

Crap. Thanks for pointing that out, I've been planning to change our `lazy_static` uses in cairo-rs and would have brought problems down the line. I think I'll subscribe to the tracking issue to see if that changes.

tdelabro:
superseded by #1262

tdelabro:
superseded by https://github.com/starkware-libs/blockifier/pull/1262

PR #478, opened 2023-05-16
Add papyrus reader class cache

giladchase:
While there is already a cache like this in `CachedState`, it cannot persist in between blocks, whereas this cache is saved on the reader, which is currently saved on the transaction executor instance.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/478)

### DIFF
--- a/crates/native_blockifier/src/papyrus_state.rs
+++ b/crates/native_blockifier/src/papyrus_state.rs
@@ -1,3 +1,6 @@
+use std::collections::hash_map::Entry;
+use std::collections::HashMap;
+
 use blockifier::execution::contract_class::{ContractClass, ContractClassV0};
 use blockifier::state::errors::StateError;
 use blockifier::state::state_api::{StateReader, StateResult};
@@ -11,17 +14,23 @@ use starknet_api::state::{StateNumber, StorageKey};
 #[path = "papyrus_state_test.rs"]
 mod test;
 
+pub type ClassCache = HashMap<ClassHash, ContractClass>;
 type RawPapyrusStateReader<'env> = papyrus_storage::state::StateReader<'env, RO>;
 
 pub struct PapyrusStateReader<'env> {
     pub reader: RawPapyrusStateReader<'env>,
+    pub contract_class_cache: ClassCache,
     // Invariant: Read-Only.
     latest_block: BlockNumber,
 }
 
 impl<'env> PapyrusStateReader<'env> {
-    pub fn new(reader: RawPapyrusStateReader<'env>, latest_block: BlockNumber) -> Self {
-        Self { reader, latest_block }
+    pub fn new(
+        reader: RawPapyrusStateReader<'env>,
+        latest_block: BlockNumber,
+        contract_class_cache: ClassCache,
+    ) -> Self {
+        Self { reader, latest_block, contract_class_cache }
     }
 
     pub fn latest_block(&self) -> &BlockNumber {
@@ -64,12 +73,19 @@ impl<'env> StateReader for PapyrusStateReader<'env> {
         class_hash: &ClassHash,
     ) -> StateResult<ContractClass> {
         let state_number = StateNumber(*self.latest_block());
-        match self.reader.get_deprecated_class_definition_at(state_number, class_hash) {
-            Ok(Some(starknet_api_contract_class)) => {
-                Ok(ContractClassV0::try_from(starknet_api_contract_class)?.into())
+        match self.contract_class_cache.entry(*class_hash) {
+            Entry::Occupied(contract_class) => Ok(contract_class.get().clone()),
+            Entry::Vacant(entry) => {
+                match self.reader.get_deprecated_class_definition_at(state_number, class_hash) {
+                    Ok(Some(starknet_api_contract_class)) => {
+                        let contract_class =
+                            ContractClassV0::try_from(starknet_api_contract_class)?.into();
+                        Ok(entry.insert(contract_class).clone())
+                    }
+                    Ok(None) => Err(StateError::UndeclaredClassHash(*class_hash)),
+                    Err(err) => Err(StateError::StateReadError(err.to_string())),
+                }
             }
-            Ok(None) => Err(StateError::UndeclaredClassHash(*class_hash)),
-            Err(err) => Err(StateError::StateReadError(err.to_string())),
         }
     }
 
--- a/crates/native_blockifier/src/papyrus_state_test.rs
+++ b/crates/native_blockifier/src/papyrus_state_test.rs
@@ -1,3 +1,5 @@
+use std::collections::HashMap;
+
 use blockifier::abi::abi_utils::selector_from_name;
 use blockifier::execution::entry_point::{CallEntryPoint, CallExecution, Retdata};
 use blockifier::retdata;
@@ -42,7 +44,8 @@ fn test_entry_point_with_papyrus_state() -> papyrus_storage::StorageResult<()> {
 
     // BlockNumber is 1 due to the initialization step above.
     let block_number = BlockNumber(1);
-    let papyrus_reader = PapyrusStateReader::new(state_reader, block_number);
+    let contract_class_cache = HashMap::new();
+    let papyrus_reader = PapyrusStateReader::new(state_reader, block_number, contract_class_cache);
     let mut state = CachedState::new(papyrus_reader);
 
     // Call entrypoint that want to write to storage, which updates the cached state's write cache.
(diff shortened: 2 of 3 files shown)


giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/478)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @elintul and @noaov1)

___
*[`crates/native_blockifier/src/papyrus_state.rs` line 17 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/478#-NVZjURe-FZf10zPxd2g:-NVZjURe-FZf10zPxd2h:b-hpm2t3) ([raw file](https://github.com/starkware-libs/blockifier/blob/2b6cfa8cba6b58d54cb168c2f3466a988c0f3175/crates/native_blockifier/src/papyrus_state.rs#L17)):*
> ```XML
> mod test;
> 
> pub type ClassCache = HashMap<ClassHash, ContractClass>;
> ```

Wasn't sure if i should make this depend on types defined in CachedState, or if the CachedState cache is still relevant if we cache things at this level.
WDYT?



_Code quote:_
```XML
pub type ClassCache = HashMap<ClassHash, ContractClass>;
```

elintul (changes requested):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/478)* status: all files reviewed, 3 unresolved discussions (waiting on @giladchase and @noaov1)

___
*[`crates/native_blockifier/src/papyrus_state.rs` line 17 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/478#-NVtW4ec-EGx7s0kkoVz:-NVtW4ec-EGx7s0kkoW-:undefined) ([raw file](https://github.com/starkware-libs/blockifier/blob/2b6cfa8cba6b58d54cb168c2f3466a988c0f3175/crates/native_blockifier/src/papyrus_state.rs#L17)):*
> ```XML
> mod test;
> 
> pub type ClassCache = HashMap<ClassHash, ContractClass>;
> ```





_Suggestion:_
```XML
ContractClassCache
```
___
*[`crates/native_blockifier/src/papyrus_state.rs` line 17 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/478#-NVZjURe-FZf10zPxd2g:-NVtWSIw772b3TGlyJVg:b-x28etf) ([raw file](https://github.com/starkware-libs/blockifier/blob/2b6cfa8cba6b58d54cb168c2f3466a988c0f3175/crates/native_blockifier/src/papyrus_state.rs#L17)):*

Wasn't sure if i should make this depend on types defined in CachedState, or if the CachedState cache is still relevant if we cache things at this level.
WDYT?
</blockquote></details>

Can you remind me what stage in the execution this caching is optimizing?
The cache we currently have in the `CachedState` is a cache in the _block level_ (meaning, also between transactions). Can you remind me why wasn't this enough when you analyzed performance?
___
*[`crates/native_blockifier/src/papyrus_state.rs` line 76 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/478#-NVtWKBW6eZRkJZZ4ToJ:-NVtWKBW6eZRkJZZ4ToK:byzjid4) ([raw file](https://github.com/starkware-libs/blockifier/blob/2b6cfa8cba6b58d54cb168c2f3466a988c0f3175/crates/native_blockifier/src/papyrus_state.rs#L76)):*
> ```XML
>     ) -> StateResult<ContractClass> {
>         let state_number = StateNumber(*self.latest_block());
>         match self.contract_class_cache.entry(*class_hash) {
> ```

Why not use `get`?



_Code quote:_
```XML
entry
```

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/478)* status: all files reviewed, 2 unresolved discussions (waiting on @elintul and @noaov1)

___
*[`crates/native_blockifier/src/papyrus_state.rs` line 17 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/478#-NVZjURe-FZf10zPxd2g:-NVwzkv-BeFXclRs-rdJ:bk8f2jw) ([raw file](https://github.com/starkware-libs/blockifier/blob/2b6cfa8cba6b58d54cb168c2f3466a988c0f3175/crates/native_blockifier/src/papyrus_state.rs#L17)):*

Can you remind me what stage in the execution this caching is optimizing?
The cache we currently have in the `CachedState` is a cache in the _block level_ (meaning, also between transactions). Can you remind me why wasn't this enough when you analyzed performance?
</blockquote></details>

Between blocks.
If two different blocks need the same contract, this PR will make the second block not have to hit the DB.
___
*[`crates/native_blockifier/src/papyrus_state.rs` line 76 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/478#-NVtWKBW6eZRkJZZ4ToJ:-NVx506m8Yo9ykYavW9L:b9b6w52) ([raw file](https://github.com/starkware-libs/blockifier/blob/2b6cfa8cba6b58d54cb168c2f3466a988c0f3175/crates/native_blockifier/src/papyrus_state.rs#L76)):*

Why not use `get`?
</blockquote></details>

tl;dr: Not so much about the `get` as it is about the `insert`; using `entry` for both saves an extra `get` query during the insert and reduces coupling.

tl (sorry :grimacing:):
Regular `insert` doesn't return the inserted value (it returns `None` in this case), meaning we'd have to add an extra `get` after the `insert` (which is coupled with the outer `get` and the `insert`, they all depend on the same `class_hash` variable).
```rust
        match self.contract_class_cache.get(class_hash) {
            Some(contract_class) => Ok(contract_class.clone()),
            None => {
                match self.reader.get_deprecated_class_definition_at(state_number, class_hash) {
                    Ok(Some(starknet_api_contract_class)) => {
                        let contract_class =
                            ContractClassV0::try_from(starknet_api_contract_class)?.into();
                        self.contract_class_cache.insert(*class_hash, contract_class);  // returns None
                        Ok(self.contract_class_cache.get(class_hash).cloned().unwrap())  // extra `get`
                    }
                    ...
```

The "insert and return value" functionality [is provided by the `entry` API](https://github.com/rust-lang/rust/issues/21644).

Note: we _could_ technically use the `get` in the outer match, and an `entry(class_hash).or_insert(contract_class)` in the inner match, like the following, but this again does an extra query + coupling 
```rust
        match self.contract_class_cache.get(class_hash) {
            Some(contract_class) => Ok(contract_class.clone()),
            None => {
                match self.reader.get_deprecated_class_definition_at(state_number, class_hash) {
                    Ok(Some(starknet_api_contract_class)) => {
                        let contract_class =
                            ContractClassV0::try_from(starknet_api_contract_class)?.into();
                        Ok(self
                            .contract_class_cache
                            .entry(*class_hash)
                            .or_insert(contract_class)
                            .clone())
                    }
```

elintul (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/478)* status: all files reviewed, 2 unresolved discussions (waiting on @giladchase and @noaov1)

___
*[`crates/native_blockifier/src/papyrus_state.rs` line 17 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/478#-NVZjURe-FZf10zPxd2g:-NVxxNmW43WidUASnfUY:b-bktc4h) ([raw file](https://github.com/starkware-libs/blockifier/blob/2b6cfa8cba6b58d54cb168c2f3466a988c0f3175/crates/native_blockifier/src/papyrus_state.rs#L17)):*

Between blocks.
If two different blocks need the same contract, this PR will make the second block not have to hit the DB.
</blockquote></details>

How will it cache information between blocks? The executor is recreated on every batch creation attempt, right?

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/478)* status: all files reviewed, 2 unresolved discussions (waiting on @elintul and @noaov1)

___
*[`crates/native_blockifier/src/papyrus_state.rs` line 17 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/478#-NVZjURe-FZf10zPxd2g:-NVy90Tj1a50WJfHNAhp:b-k042y5) ([raw file](https://github.com/starkware-libs/blockifier/blob/2b6cfa8cba6b58d54cb168c2f3466a988c0f3175/crates/native_blockifier/src/papyrus_state.rs#L17)):*

How will it cache information between blocks? The executor is recreated on every batch creation attempt, right?
</blockquote></details>

Ya, i guess i was assuming it'd be possible to save this cache on the python side (on the Batcher instance?) and then pass it back to rust during the initialization of the executor every batch?

Now that I think about it, this won't be totally efficient because the (big) cache will be re-serialized every block, but that can be fixed if I make the cache be a mutable reference to the python cache instance.

giladchase:
Closing, this is wrong.

The executor dies between blocks, so no point in saving stuff on the reader, it won't persist.
The only way of making it persist is to pass it back to Python, which incurs heavy serialization costs.

PR #518, opened 2023-05-23
Move papyrus implementation of State trait to papyrus

dan-starkware:


giladchase (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/518)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @dan-starkware)

___
*[`crates/native_blockifier/Cargo.toml` line 17 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/518#-NWCKyhU9r9Fi3BAjzIY:-NWCKyhU9r9Fi3BAjzIZ:b-19l7jt) ([raw file](https://github.com/starkware-libs/blockifier/blob/cf4a82c2725e8f7b693b2280b1b4a986c4c7470a/crates/native_blockifier/Cargo.toml#L17)):*
> ```TOML
> [dependencies]
> # TODO: use path once we move to a mono-repo.
> blockifier = { git = "https://github.com/starkware-libs/blockifier", branch = "main-v0.12.0" }
> ```

I think this will slow us down too much, we need to find a better way to do this.

If we merge this, it'll be very difficult (and in most cases impossible) to simultaneously work on blockifier/native_blockifier (locally I mean).

For example, if I add a new struct to blockifier that native_blockifier wants to import and use, I'll have first to get that struct _merged into main_, before I'll even be able to _test_ out its behaviour locally with native_blockifier.

I looked into `patch`, and seems like it doesn't cover the case we want.
I also looked at what rust-analyzer does, and seems like they use this notation, which looks hackish and possibly what we want:
```rust
// An example for a dependency in rust-analyzer
base-db = { path = "./crates/base-db", version = "0.0.0" }
```
IIUC, this looks in local path first, and if that doesn't exist falls back on version 0.0.0 in cratesio?



_Code quote:_
```TOML
blockifier = { git = "https://github.com/starkware-libs/blockifier", branch = "main-v0.12.0" }
```

dan-starkware:
*[`crates/native_blockifier/Cargo.toml` line 17 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/518#-NWCKyhU9r9Fi3BAjzIY:-NWCcZhx90zec5lNQpUc:bauqae9) ([raw file](https://github.com/starkware-libs/blockifier/blob/cf4a82c2725e8f7b693b2280b1b4a986c4c7470a/crates/native_blockifier/Cargo.toml#L17)):*

I think this will slow us down too much, we need to find a better way to do this.

If we merge this, it'll be very difficult (and in most cases impossible) to simultaneously work on blockifier/native_blockifier (locally I mean).

For example, if I add a new struct to blockifier that native_blockifier wants to import and use, I'll have first to get that struct _merged into main_, before I'll even be able to _test_ out its behaviour locally with native_blockifier.

I looked into `patch`, and seems like it doesn't cover the case we want.
I also looked at what rust-analyzer does, and seems like they use this notation, which looks hackish and possibly what we want:
```rust
// An example for a dependency in rust-analyzer
base-db = { path = "./crates/base-db", version = "0.0.0" }
```
IIUC, this looks in local path first, and if that doesn't exist falls back on version 0.0.0 in cratesio?
</blockquote></details>

We can either publish the crate or unite the repos. What do you suggest we do in the meantime?

PR #585, opened 2023-06-08
Wrap `PyTransactionExecutor` in a newtype

giladchase:
This is done in order to have a convenient `close` function, which deallocates the executor.

Without this, the executor instance isn't deallocated (at least not immediately), which causes READERS_FULL errors from MDBX.

Python version bump: https://github.com/starkware-industries/starkware/pull/30334

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/585)

### DIFF
--- a/crates/native_blockifier/src/py_transaction.rs
+++ b/crates/native_blockifier/src/py_transaction.rs
@@ -279,11 +279,68 @@ pub fn py_tx(
     }
 }
 
+/// Wraps the transaction executor in an optional, to allow an explicit deallocation of it.
+/// The explicit deallocation is needed since PyO3 can't track lifetimes within Python.
 #[pyclass]
+pub struct PyTransactionExecutor {
+    pub executor: Option<PyTransactionExecutorInner>,
+}
+
+#[pymethods]
+impl PyTransactionExecutor {
+    #[new]
+    #[args(general_config, block_info, papyrus_storage)]
+    pub fn create(
+        general_config: &PyAny,
+        block_info: &PyAny,
+        papyrus_storage: &Storage,
+    ) -> NativeBlockifierResult<Self> {
+        log::debug!("Initializing Transaction Executor...");
+        let executor =
+            PyTransactionExecutorInner::create(general_config, block_info, papyrus_storage)?;
+        log::debug!("Initialized Transaction Executor.");
+
+        Ok(Self { executor: Some(executor) })
+    }
+
+    #[args(tx, raw_contract_class, enough_room_for_tx)]
+    pub fn execute(
+        &mut self,
+        tx: &PyAny,
+        raw_contract_class: Option<&str>,
+        // This is functools.partial(bouncer.add, tw_written=tx_written).
+        enough_room_for_tx: &PyAny,
+    ) -> NativeBlockifierResult<(
+        Py<PyTransactionExecutionInfo>,
+        HashMap<PyFelt, PyContractClassSizes>,
+    )> {
+        self.executor().execute(tx, raw_contract_class, enough_room_for_tx)
+    }
+
+    pub fn finalize(&mut self) -> PyStateDiff {
+        log::debug!("Finalizing execution...");
+        let state_diff = self.executor().finalize();
+        self.close();
+        log::debug!("Finalized execution.");
+
+        state_diff
+    }
+
+    pub fn close(&mut self) {
+        self.executor = None;
+    }
+}
+
+impl PyTransactionExecutor {
+    fn executor(&mut self) -> &mut PyTransactionExecutorInner {
+        self.executor.as_mut().expect("Transaction executor should be initialized.")
+    }
+}
+
 // To access a field you must use `self.borrow_{field_name}()`.
 // Alternately, you can borrow the whole object using `self.with[_mut]()`.
 #[ouroboros::self_referencing]
-pub struct PyTransactionExecutor {
+pub struct PyTransactionExecutorInner {
     pub block_context: BlockContext,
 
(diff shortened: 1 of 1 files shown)


elintul (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/585)* status: all files reviewed, 2 unresolved discussions (waiting on @giladchase)

___
*[`crates/native_blockifier/src/py_transaction.rs` line 283 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/585#-NXQtrLt2LOjstkVABa7:-NXQtrLt2LOjstkVABa8:boxr521) ([raw file](https://github.com/starkware-libs/blockifier/blob/4b58e697316a6ff208dc3d07787092844e80d1b0/crates/native_blockifier/src/py_transaction.rs#L283)):*
> ```XML
> 
> /// Wraps the transaction executor in an optional, to allow an explicit deallocation of it.
> /// The explicit deallocation is needed since PYO3 can't track lifetimes within Python.
> ```

Can you explain this sentence?



_Suggestion:_
```XML
/// The explicit deallocation is needed since PyO3 can't track lifetimes within Python
```
___
*[`crates/native_blockifier/src/py_transaction.rs` line 340 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/585#-NXQuC045e29rWiy9lGo:-NXQuC045e29rWiy9lGp:b-yrvoer) ([raw file](https://github.com/starkware-libs/blockifier/blob/4b58e697316a6ff208dc3d07787092844e80d1b0/crates/native_blockifier/src/py_transaction.rs#L340)):*
> ```XML
> }
> 
> #[pyclass]
> ```

Still needed?



_Code quote:_
```XML
#[pyclass]
```

giladchase (commented):
We tried it when we had a similar error with the `Storage` instance; didn't seem to work.

Maybe PYO3 doesn’t deallocate cause it doesn’t maintain a ref-counter for Python objects, so a single `del` of a ref doesn’t trigger a `drop`?

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/585#-:-NXeNACqAOemy4O8V2gP:b-woqei7)* status: all files reviewed, 2 unresolved discussions (waiting on @giladchase)

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/585)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @elintul)

___
*[`crates/native_blockifier/src/py_transaction.rs` line 283 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/585#-NXQtrLt2LOjstkVABa7:-NXeQAWy-bGg0FBjPaGv:b-nxe0vt) ([raw file](https://github.com/starkware-libs/blockifier/blob/4b58e697316a6ff208dc3d07787092844e80d1b0/crates/native_blockifier/src/py_transaction.rs#L283)):*

Can you explain this sentence?
</blockquote></details>

Context: 
Generally, Rust drops objects when their lifetime ends.
For example, if an instance is _moved_ into a function and not returned, it is deallocated by the compiler after the last line that uses it (the compiler calls `mem::drop(object)` which is like `free`).

With PyO3:
When context switching to Python, the compiler can't track anything, so it deallocates things rather poorly (we still don't fully understand the deallocation logic).
However, if the deallocation is explicit in Rust, it is deallocated as usual.

We needed a `newtype` for this, because we are deallocating the entire Executor object, so this can't be a method (an object can't free itself :smiling_face_with_tear: it'll lead to instance holding it becoming undefined).
___
*[`crates/native_blockifier/src/py_transaction.rs` line 340 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/585#-NXQuC045e29rWiy9lGo:-NXeSDLc2jH1DRZNi45a:bibkq7i) ([raw file](https://github.com/starkware-libs/blockifier/blob/4b58e697316a6ff208dc3d07787092844e80d1b0/crates/native_blockifier/src/py_transaction.rs#L340)):*

Still needed?
</blockquote></details>

Nope yr right.
I think i thought this was needed since the object is part of an object passed into Python, but it isn't needed since Python can't access it (directly).

elintul (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/585#-:-NXeY88L66aV2vKRFrwH:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @giladchase)

___
*[`crates/native_blockifier/src/py_transaction.rs` line 283 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/585#-NXQtrLt2LOjstkVABa7:-NXeY21a7uK_y2ixby87:b-ucfknh) ([raw file](https://github.com/starkware-libs/blockifier/blob/4b58e697316a6ff208dc3d07787092844e80d1b0/crates/native_blockifier/src/py_transaction.rs#L283)):*

Context: 
Generally, Rust drops objects when their lifetime ends.
For example, if an instance is _moved_ into a function and not returned, it is deallocated by the compiler after the last line that uses it (the compiler calls `mem::drop(object)` which is like `free`).

With PyO3:
When context switching to Python, the compiler can't track anything, so it deallocates things rather poorly (we still don't fully understand the deallocation logic).
However, if the deallocation is explicit in Rust, it is deallocated as usual.

We needed a `newtype` for this, because we are deallocating the entire Executor object, so this can't be a method (an object can't free itself :smiling_face_with_tear: it'll lead to instance holding it becoming undefined).
</blockquote></details>

Thanks!

PR #638, opened 2023-06-20
Add global contract class cache

giladchase:
The idea: the cache should live across blocks, and not perform copies or serialization.

Assumption 1: Only places that can own it across blocks is the `Storage` instance. Assumption 2: CachedState needs to own a reference to the cache to use in `get_compiled_contract_class`.

Implementation constraint: Even though cached state only needs immutable access to the cache, when getting to the type `CachedState<MutRefState>` the compiler couldn't quite handle an immutable reference (or Arc, for simplicity) as a field in `CachedState`.

Why `Arc<RwLock>`: Given the implementation constraint above, it seems like [multiple of multiple owners is needed](https://doc.rust-lang.org/book/ch15-05-interior-mutability.html#having-multiple-owners-of-mutable-data-by-combining-rct-and-refcellt), that is, `Arc<RefCell>`. However, `RefCell` is not thread-safe, so [Pyo3 won't allow it to be part of a `#pyclass`](https://pyo3.rs/v0.19.0/migration.html?highlight=refcell#pyclass-structs-must-now-be-send-or-unsendable). To make `Arc<RefCell>` thread-safe, we can use `Arc<Mutex>`, or, to communicate that reads are much more common than writes, `Arc<RwLock>`.

On using Arc<RwLock> as a thread-safe version of the Arc<RefCell> pattern

Question: Should the cache have bounded size?


elintul (commented):
> Reviewed 3 of 13 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/638)* status: 3 of 13 files reviewed, all discussions resolved (waiting on @Yoni-Starkware)

PR #835, opened 2023-08-16
Replace *py_deploy_account* method with PyDeployAccountTransaction struct.

NirLevi-starkware:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/835)

### DIFF
--- a/crates/native_blockifier/src/py_transaction.rs
+++ b/crates/native_blockifier/src/py_transaction.rs
@@ -106,25 +106,45 @@ pub fn py_declare(
     Ok(DeclareTransaction::new(sn_api_tx, account_data_context.transaction_hash, contract_class)?)
 }
 
-pub fn py_deploy_account(tx: &PyAny) -> NativeBlockifierResult<DeployAccountTransaction> {
-    let account_data_context = py_account_data_context(tx)?;
+#[derive(FromPyObject)]
+pub struct PyDeployAccountTransaction {
+    pub nonce: PyFelt,
+    pub hash_value: PyFelt,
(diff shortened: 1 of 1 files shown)


giladchase (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/835)* status: all files reviewed, 1 unresolved discussion (waiting on @NirLevi-starkware)

___
*[`crates/native_blockifier/src/py_transaction.rs` line 125 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/835#-Nc1HjMT6Mi9Kw6fitwx:-Nc1HjMT6Mi9Kw6fitwy:b-ng4kw0) ([raw file](https://github.com/starkware-libs/blockifier/blob/fbb09765daf7c92479f46bb5b04781c0229a8a9a/crates/native_blockifier/src/py_transaction.rs#L125)):*
> ```XML
>     fn from(py_deploy_account_tx: PyDeployAccountTransaction) -> Self {
>         let tx = starknet_api::transaction::DeployAccountTransaction {
>             max_fee: Fee(u128::try_from(py_deploy_account_tx.max_fee).unwrap()),
> ```

Generally we don't use `unwrap`, only in tests, or when the `unwrap` can't fail for sure, and it is absolutely obvious why.

Sometimes we use `expect`, but that's only in very specific cases where we want to crash the service.

In particular, `from`'s should never fail, so if you find yourself wanting to unwrap, odds are you need to impl `TryFrom` rather than `From`, as is the case here.



_Suggestion:_
```XML
?
```

PR #838, opened 2023-08-17
Refactor cache handling in `CachedState`

giladchase:
- Move state cache-updating logic into a method `update_cache`. This will soon be used when we add an additional transactional state type. Consequently, `StateCache` is now public (since `update_cache` is part of the public API).
- Move local and global contract cache into a method `update_contract_class_caches`, for the same reason as the above.
- Code convention fix: move the private method `update_initial_values_of_write_only_access` to the bottom of the `impl` block.
This is why `move_classes_to_global_cache` and `global_class_hash_to_class` appear in the diff, they were hoisted up above `update_initial_values_of_write_only_access`(but were not changed).

Context: getting ready for removing the python callback in `native_blockifier`. 
This will require a different type of `TransactionalState` (without a lifetime, because of Pyo3), so the current change prevents duplicated code in its `commit` stage.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/838)

### DIFF
--- a/crates/blockifier/src/state/cached_state.rs
+++ b/crates/blockifier/src/state/cached_state.rs
@@ -98,6 +98,40 @@ impl<S: StateReader> CachedState<S> {
         })
     }
 
+    /// Drains contract-class cache collected during execution and updates the global cache.
+    pub fn move_classes_to_global_cache(&mut self) {
+        let contract_class_updates: Vec<_> = self.class_hash_to_class.drain().collect();
+        for (key, value) in contract_class_updates {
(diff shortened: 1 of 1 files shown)


elintul (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/838#-:-Nc2bPnDEho2VRrPHUM1:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @noaov1)

PR #845, opened 2023-08-22
Implement TryFrom for DataAvailabilityMode enum in order to be able t…

barak-b-starkware:
…o convert a StarkFelt into an enum variant.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/845)

### DIFF
--- a/crates/blockifier/src/execution/syscalls/hint_processor.rs
+++ b/crates/blockifier/src/execution/syscalls/hint_processor.rs
@@ -45,7 +45,7 @@ use crate::execution::syscalls::{
     SyscallResponse, SyscallResponseWrapper, SyscallResult, SyscallSelector,
 };
 use crate::state::errors::StateError;
-use crate::state::state_api::{DataAvailabilityMode, State};
+use crate::state::state_api::State;
 use crate::transaction::transaction_utils::update_remaining_gas;
 
 pub type SyscallCounter = HashMap<SyscallSelector, usize>;
--- a/crates/blockifier/src/execution/syscalls/hint_processor.rs
+++ b/crates/blockifier/src/execution/syscalls/hint_processor.rs
@@ -45,7 +45,7 @@ use crate::execution::syscalls::{
     SyscallResponse, SyscallResponseWrapper, SyscallResult, SyscallSelector,
 };
 use crate::state::errors::StateError;
-use crate::state::state_api::State;
+use crate::state::state_api::{DataAvailabilityMode, State};
 use crate::transaction::transaction_utils::update_remaining_gas;
 
 pub type SyscallCounter = HashMap<SyscallSelector, usize>;
--- a/crates/blockifier/src/execution/syscalls/mod.rs
+++ b/crates/blockifier/src/execution/syscalls/mod.rs
@@ -28,6 +28,7 @@ use crate::execution::execution_utils::{
     write_felt, write_maybe_relocatable, write_stark_felt, ReadOnlySegment,
 };
 use crate::execution::syscalls::hint_processor::{INVALID_INPUT_LENGTH_ERROR, OUT_OF_GAS_ERROR};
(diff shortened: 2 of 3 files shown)


noaov1 (changes requested):
> Reviewed 1 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/845)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on @barak-b-starkware)

___
*[`crates/blockifier/src/execution/syscalls/mod.rs` line 107 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/845#-NcStK6n3Ub7IZoQ5WW3:-NcVui8E82E5DkKEQddw:b-iz4tim) ([raw file](https://github.com/starkware-libs/blockifier/blob/3613ad375ce88a1b38db5d7b54ee0aa590aac630/crates/blockifier/src/execution/syscalls/mod.rs#L107)):*

I’m not sure about this `impl`, and I have some questions:

1.  Would you choose another trait to implement to get this behavior?
2.  When we’ll move the enum to the starknet_api repo, I would like to leave the `impl` block in the blockifier repo since those are inner details (how we convert the syscall output into this enum), so I placed it in this `syscalls/mod.rs` file, WDYT?
3.  I didn’t use the `match` mechanism and preferred the `if...else...` here. I had difficulty using variables in the `match` arms. Is this the right way to handle it?
4. Creating the `zero` and `one` seems overkill, but I’m unfamiliar with any other way to do that. I saw that the `one` is implemented in the starknet_api repo but is still unavailable in the blockifier repo. WDYT?
5.  What about the use of `unwrap`? I remember it’s an issue. What is our policy?
</blockquote></details>

1. I think it's good. Can you please refer me to the cairo code that sends it? Is it a felt?
2. I think it's better to have it in SN API. BTW- you cannot implement an external trait for an external enum, so it will not be possible.
3. It is okay :)
4. See my comments below.
5. No `unwrap()`s. This can lead to panic in the code without proper notification. If it’s something you're 100% sure will work (unless there’s a bug in the code), then you should use `expect()`. Otherwise, you must return an error. (that’s my thoughts, anyway).
___
*[`crates/blockifier/src/execution/syscalls/mod.rs` line 115 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/845#-NcWDaMm5GbLFbYfTWdz:-NcWDaMm5GbLFbYfTWe-:undefined) ([raw file](https://github.com/starkware-libs/blockifier/blob/3613ad375ce88a1b38db5d7b54ee0aa590aac630/crates/blockifier/src/execution/syscalls/mod.rs#L115)):*
> ```XML
>             0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0,
>         ])
>         .unwrap();
> ```





_Suggestion:_
```XML
        let zero = StarkFelt::from(0_u8);
```
___
*[`crates/blockifier/src/execution/syscalls/mod.rs` line 120 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/845#-NcWDeb71yVGQBm-3jpE:-NcWDeb88PiBCQDuH2Iz:undefined) ([raw file](https://github.com/starkware-libs/blockifier/blob/3613ad375ce88a1b38db5d7b54ee0aa590aac630/crates/blockifier/src/execution/syscalls/mod.rs#L120)):*
> ```XML
>             0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x1,
>         ])
>         .unwrap();
> ```





_Suggestion:_
```XML
        let one = StarkFelt::from(1_u8);
```

PR #1039, opened 2023-10-26
Add l1 support to `ActualCost`

giladchase:
- Calculate l1 actual fee via `ActualCost`: its logic was almost identical to the logic of `calculate_actual_fee_and_resources`.
- Required adding `l1_payload_size` which was always None for `AccountTransaction`, which were the only txs `ActualCost` supported thus far.
- Also added `sender_address` for the opposite reason as the prev bullet.
- Removed NEW_TOKEN_SUPPORT TODO since it is already fixed inside `ActualCost`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1039)

### DIFF
--- a/crates/blockifier/src/fee/actual_cost.rs
+++ b/crates/blockifier/src/fee/actual_cost.rs
@@ -1,5 +1,6 @@
 use std::cmp::min;
 
+use starknet_api::core::ContractAddress;
 use starknet_api::transaction::Fee;
 
 use crate::abi::constants as abi_constants;
@@ -21,6 +22,18 @@ pub struct ActualCost {
     pub actual_resources: ResourcesMapping,
 }
 
+impl ActualCost {
+    pub fn builder_for_l1_handler(
+        block_context: &BlockContext,
+        tx_context: AccountTransactionContext,
+        l1_handler_payload_size: usize,
+    ) -> ActualCostBuilder<'_> {
+        ActualCostBuilder::new(block_context, tx_context, TransactionType::L1Handler)
+            .without_sender_address()
+            .with_l1_payload_size(l1_handler_payload_size)
+    }
+}
+
 #[derive(Debug, Clone)]
 // Invariant: private fields initialized after `new` is called via dedicated methods.
 pub struct ActualCostBuilder<'a> {
@@ -30,10 +43,14 @@ pub struct ActualCostBuilder<'a> {
     validate_call_info: Option<&'a CallInfo>,
     execute_call_info: Option<&'a CallInfo>,
     state_changes: StateChanges,
+    sender_address: Option<ContractAddress>,
+    l1_payload_size: Option<usize>,
+    is_reverted: bool,
+    n_reverted_steps: usize,
 }
 
 impl<'a> ActualCostBuilder<'a> {
-    // Recommendation: use constructor from account transaction to build this.
+    // Recommendation: use constructor from account transaction, or from actual cost, to build this.
     pub fn new(
         block_context: &BlockContext,
         account_tx_context: AccountTransactionContext,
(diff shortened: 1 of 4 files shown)


dorimedini-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1039)* status: all files reviewed, 1 unresolved discussion (waiting on @elintul and @giladchase)

*[a discussion](https://reviewable.io/reviews/starkware-libs/blockifier/1039#-Nhf-icH03oJrsrHQKAS:-Nhf-icH03oJrsrHQKAT:btkd8oo) (no related file):*
Two remarks:
1. Please add a L1 handler test with fee comparison, and then run it before & after this change. If such a test already exists - please point it out?
2. I don't really like the fine-grained control you give the user of ActualCostBuilder. I don't want to know that in the case of L1 handler, I must ignore the sender address and add explicit L1 usage. Please implement a `build_for_l1_handler` method on the cost builder, which accepts L1 data length as input, and knows to ignore the sender address... no point in extra fields and "without_sender" and "with_l1_data" methods

___

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1039)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @dorimedini-starkware and @elintul)

*[a discussion](https://reviewable.io/reviews/starkware-libs/blockifier/1039#-Nhf-icH03oJrsrHQKAS:-NiDRkJt7GYGORhPj0Kj:b-g0x4kc) (no related file):*

Two remarks:
1. Please add a L1 handler test with fee comparison, and then run it before & after this change. If such a test already exists - please point it out?
2. I don't really like the fine-grained control you give the user of ActualCostBuilder. I don't want to know that in the case of L1 handler, I must ignore the sender address and add explicit L1 usage. Please implement a `build_for_l1_handler` method on the cost builder, which accepts L1 data length as input, and knows to ignore the sender address... no point in extra fields and "without_sender" and "with_l1_data" methods
</blockquote></details>

1) IIRC we don't charge yet for l1 handlers or something like that? the code also suggests that is indeed the case, see [this](https://github.com/starkware-libs/blockifier/blob/295147f7081ce6dab0eca1ecb773a3ed5d4abc98/crates/blockifier/src/transaction/transaction_execution.rs#L105-L109) and right after it we have [this](https://github.com/starkware-libs/blockifier/blob/295147f7081ce6dab0eca1ecb773a3ed5d4abc98/crates/blockifier/src/transaction/transaction_execution.rs#L115). I'll ask around and update here.
2) i see your point, i added the constructor with all the special stuff to `ActualCost` though, if that's OK. 
Rationale: the builder in the builder pattern is expected to be stupid and very fine-grained, so i added the smarter stuff to the struct, which doesn't have any special assumptions.

To be more consistent with the above, i changed the builder API to only have one `build` command, since having two `build`s is also abusing the pattern a bit (where a builder should be a bunch of setters + a single `build` command)--- WDYT? I did so under the assumption that this doesn't increase the granularity at account_transaction.rs :sweat_smile: 

___

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1039)* status: 2 of 4 files reviewed, 1 unresolved discussion (waiting on @dorimedini-starkware and @elintul)

___
*[`crates/blockifier/src/transaction/transaction_execution.rs` line 99 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1039#-NiEI0rW-Hv0FDrEpPq8:-NiEfxYuBdMH0az_Cym2:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/295147f7081ce6dab0eca1ecb773a3ed5d4abc98/crates/blockifier/src/transaction/transaction_execution.rs#L99)):*

please rename, less confusing
</blockquote></details>

Done.

dorimedini-starkware (approved):
:lgtm:

> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1039#-:-NiFNCUG2nlJsMWlPZ2L:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @elintul)

PR #1149, opened 2023-11-19
Make `native_blockifier.Storage` testable + 1 use

giladchase:
- Traitify `Storage`, so that it can be mocked as `FakeStorage`.
- Previous `Storage` became `PapyrusStorage` and methods moved to `impl trait`, no logic changes.
- Add test for get_block_id: tests the state the system is when loading from a `papyrus` snapshot, which will return a `StarkHash` (this hit a bug in the past, when we assumed it was `u64`)

Python: https://reviewable.io/reviews/starkware-industries/starkware/32665

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1149)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2110,6 +2110,7 @@ dependencies = [
  "log",
  "num-bigint",
  "papyrus_storage",
+ "pretty_assertions",
  "pyo3",
  "pyo3-log",
  "serde_json",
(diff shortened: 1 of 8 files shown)


giladchase:
Whoops NVM too big, i''m closing and splitting in two

PR #1232, opened 2023-12-08
refacto: cleanup state traits, CachedState interior mutability

tdelabro:
In this PR:
- `StateReader` doesn't require a mutable reference to self anymore
- therefore, `CachedState` uses a `Cell` to have interior mutability for it's cache
- All state traits arguments are given by value, not ref, as they all impl `Copy`
- All state traits methods return a `StateResult` as we don't know for sure all impl will be infallible
- The method to compute state diff is taken out of the state traits, and into `impl CachedState`

This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1232)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,3 @@ target
 *.DS_Store
 
 tmp_venv/*
-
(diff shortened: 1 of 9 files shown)


tdelabro:
superseded by #1325

PR #1348, opened 2024-01-18
feat: add VersionedConstants

giladchase:
- Currently only covers the constants previously contained in `BlockInfo`. 
TODO: Add check if this covers all chain ids, which might have different contstants.
- Remove `BlockContextArgs`: no longer needed now that BlockContext can be initialized by 3 args of public types.

Python: https://reviewable.io/reviews/starkware-industries/starkware/33539

This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)


elintul (changes requested):
> Reviewed 6 of 24 files at r1, 1 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: 7 of 24 files reviewed, 5 unresolved discussions (waiting on @giladchase and @noaov1)

___
*[`crates/blockifier/resources/versioned_constants.json` line 1 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NoppHJW09g8MOSEMunE:-NoppHJW09g8MOSEMunF:b-n0e43) ([raw file](https://github.com/starkware-libs/blockifier/blob/4ad369687499397fbaa893dd68801ccf668f3a3d/crates/blockifier/resources/versioned_constants.json#L1)):*
> ```JSON
> {
> ```

Can you please reference the origin of these values?
___
*[`crates/blockifier/src/versioned_constants.rs` line 26 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NoppVhZ4GAkapqj93JK:-NoppVhZ4GAkapqj93JL:b-7zi8tj) ([raw file](https://github.com/starkware-libs/blockifier/blob/4ad369687499397fbaa893dd68801ccf668f3a3d/crates/blockifier/src/versioned_constants.rs#L26)):*
> ```XML
> 
> impl VersionedConstants {
>     pub fn latest_constants() -> &'static Self {
> ```

WDYM?



_Code quote:_
```XML
latest
```
___
*[`crates/blockifier/src/versioned_constants.rs` line 34 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NoppdZNDskQszz040eE:-NoppdZNDskQszz040eF:bu3xy2g) ([raw file](https://github.com/starkware-libs/blockifier/blob/4ad369687499397fbaa893dd68801ccf668f3a3d/crates/blockifier/src/versioned_constants.rs#L34)):*
> ```XML
>     type Error = VersionedConstantsError;
> 
>     fn try_from(path: &Path) -> Result<Self, Self::Error> {
> ```

Feels like a generic impl. and one that we already have; is this the case?



_Code quote:_
```XML
try_from
```
___
*[`crates/blockifier/src/versioned_constants.rs` line 35 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Noppaps1kmS8yH2K6Yv:-Noppaps1kmS8yH2K6Yw:undefined) ([raw file](https://github.com/starkware-libs/blockifier/blob/4ad369687499397fbaa893dd68801ccf668f3a3d/crates/blockifier/src/versioned_constants.rs#L35)):*
> ```XML
> 
>     fn try_from(path: &Path) -> Result<Self, Self::Error> {
>         let config_raw = std::fs::read_to_string(path)?;
> ```





_Suggestion:_
```XML
raw_constants
```
___
*[`crates/blockifier/src/versioned_constants.rs` line 41 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NoppkfO-UIB_QsqjUYM:-NoppkfO-UIB_QsqjUYN:bs7fvbp) ([raw file](https://github.com/starkware-libs/blockifier/blob/4ad369687499397fbaa893dd68801ccf668f3a3d/crates/blockifier/src/versioned_constants.rs#L41)):*
> ```XML
> 
> #[derive(Debug, Error)]
> pub enum VersionedConstantsError {
> ```

Please alphabetize; also, do we usually add a blank line between variants?



_Code quote:_
```XML
VersionedConstantsError
```

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: 7 of 24 files reviewed, 3 unresolved discussions (waiting on @elintul and @noaov1)

___
*[`crates/blockifier/resources/versioned_constants.json` line 1 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NoppHJW09g8MOSEMunE:-NopubV73C5F78dOAlqY:bg24uvd) ([raw file](https://github.com/starkware-libs/blockifier/blob/4ad369687499397fbaa893dd68801ccf668f3a3d/crates/blockifier/resources/versioned_constants.json#L1)):*

Can you please reference the origin of these values?
</blockquote></details>

Sure: you can see them in `DEFAULT_CAIRO_RESOURCE_FEE_WEIGHTS` in Python, they are the defaults.
Also in the same module you have `build_general_config` which is the one used instead of these defaults, and which calculates the exact same modules (i calculated it manually).
(Note that `build_general_config` is a bit funky, it deep_copies the argument then modifies it and returns it.)
___
*[`crates/blockifier/src/versioned_constants.rs` line 26 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NoppVhZ4GAkapqj93JK:-NopuzSH4pi6wVwa78RE:b-wflzbx) ([raw file](https://github.com/starkware-libs/blockifier/blob/4ad369687499397fbaa893dd68801ccf668f3a3d/crates/blockifier/src/versioned_constants.rs#L26)):*

WDYM?
</blockquote></details>

Probably not the best of names :sweat_smile:
 Basically in the `resources` directory you'll have one file labeld `verrsioned_constants.json` and as we progress in blockifier versions, `versioned_constants_0_13_1.json,  versioned_constants_0_13_2.json,   versioned_constants_0_14_0.json` and so on.

In particular, there will always be a single `versioned_constants.json` in that directory, which i call `latest`, and which symbolizes the constants that the current version of blockifier has shipped with. 

For example, when we publish blockifier version 1.2.3, `versioned_constants.json` will be the constants that are the "defaults" of this version.
I could have called this `default_constants_for_current_blockifier_version`, but `default` too much implies a trivial constructor in rust (with 0 values in everything), so i decided against it.

I'm not sure i explained it properly, i'll ping you offline if it's indeed the case.
___
*[`crates/blockifier/src/versioned_constants.rs` line 34 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NoppdZNDskQszz040eE:-NopwTsB0fGnOxgw73du:b-nsn1gu) ([raw file](https://github.com/starkware-libs/blockifier/blob/4ad369687499397fbaa893dd68801ccf668f3a3d/crates/blockifier/src/versioned_constants.rs#L34)):*

Feels like a generic impl. and one that we already have; is this the case?
</blockquote></details>

Can you explain a bit more what you mean plz?
I'm not sure what "feels like a generic impl" means here.
___
*[`crates/blockifier/src/versioned_constants.rs` line 41 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NoppkfO-UIB_QsqjUYM:-Nopx5ESAm7ndxSEWHA9:but1dej) ([raw file](https://github.com/starkware-libs/blockifier/blob/4ad369687499397fbaa893dd68801ccf668f3a3d/crates/blockifier/src/versioned_constants.rs#L41)):*

Please alphabetize; also, do we usually add a blank line between variants?
</blockquote></details>

Done, and no :) so i removed blanks.
I also removed an unused variant (it remained from my attempt at support multiple chains.

elintul (changes requested):
> Reviewed 15 of 24 files at r1, 2 of 3 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: all files reviewed, 8 unresolved discussions (waiting on @giladchase and @noaov1)

___
*[`crates/blockifier/resources/versioned_constants.json` line 2 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Noq2Shw1r-fzXto8eJ_:-Noq2Shw1r-fzXto8eJa:b53zxun) ([raw file](https://github.com/starkware-libs/blockifier/blob/65a3fabf063064395b8846422ce964d59dda2ce7/crates/blockifier/resources/versioned_constants.json#L2)):*
> ```JSON
> {
>     "vm_resource_fee_cost": {
> ```

Please alphabetize the builtins.



_Code quote:_
```JSON
vm_resource_fee_cost
```
___
*[`crates/blockifier/resources/versioned_constants.json` line 15 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Noq2fmm0IAgCqz6QWIh:-Noq2fmm0IAgCqz6QWIi:brondd6) ([raw file](https://github.com/starkware-libs/blockifier/blob/65a3fabf063064395b8846422ce964d59dda2ce7/crates/blockifier/resources/versioned_constants.json#L15)):*
> ```JSON
>     },
>     "invoke_tx_max_n_steps": 3000000,
>     "validate_max_n_steps": 1000000,
> ```

And these?



_Code quote:_
```JSON
    "invoke_tx_max_n_steps": 3000000,
    "validate_max_n_steps": 1000000,
```
___
*[`crates/blockifier/src/versioned_constants.rs` line 26 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NoppVhZ4GAkapqj93JK:-Noq0qzK5RMUCRrjg_ow:bk8dfsz) ([raw file](https://github.com/starkware-libs/blockifier/blob/4ad369687499397fbaa893dd68801ccf668f3a3d/crates/blockifier/src/versioned_constants.rs#L26)):*

Probably not the best of names :sweat_smile:
 Basically in the `resources` directory you'll have one file labeld `verrsioned_constants.json` and as we progress in blockifier versions, `versioned_constants_0_13_1.json,  versioned_constants_0_13_2.json,   versioned_constants_0_14_0.json` and so on.

In particular, there will always be a single `versioned_constants.json` in that directory, which i call `latest`, and which symbolizes the constants that the current version of blockifier has shipped with. 

For example, when we publish blockifier version 1.2.3, `versioned_constants.json` will be the constants that are the "defaults" of this version.
I could have called this `default_constants_for_current_blockifier_version`, but `default` too much implies a trivial constructor in rust (with 0 values in everything), so i decided against it.

I'm not sure i explained it properly, i'll ping you offline if it's indeed the case.
</blockquote></details>

I understood, consider adding a short docstring to this method.
___
*[`crates/blockifier/src/versioned_constants.rs` line 34 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NoppdZNDskQszz040eE:-Noq0wGu-rnBq45-plsX:bcdzt6r) ([raw file](https://github.com/starkware-libs/blockifier/blob/4ad369687499397fbaa893dd68801ccf668f3a3d/crates/blockifier/src/versioned_constants.rs#L34)):*

Can you explain a bit more what you mean plz?
I'm not sure what "feels like a generic impl" means here.
</blockquote></details>

Reading from a file and deserializing.
___
*[`crates/blockifier/src/versioned_constants.rs` line 12 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Noq0gNcAmzU2VweqH_1:-Noq0gNd1KmBZ36dLdpg:btlefw9) ([raw file](https://github.com/starkware-libs/blockifier/blob/65a3fabf063064395b8846422ce964d59dda2ce7/crates/blockifier/src/versioned_constants.rs#L12)):*
> ```XML
> const DEFAULT_CONSTANTS_JSON: &str = include_str!("../resources/versioned_constants.json");
> static DEFAULT_CONSTANTS: Lazy<VersionedConstants> =
>     Lazy::new(|| serde_json::from_str(DEFAULT_CONSTANTS_JSON).expect("malformed config.json"));
> ```

Please capitalize and phrase as a sentence.



_Code quote:_
```XML
malformed config.json
```
___
*[`crates/blockifier/src/execution/entry_point.rs` line 73 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Noq196875y4etQEX2WT:-Noq196875y4etQEX2WU:boke5lr) ([raw file](https://github.com/starkware-libs/blockifier/blob/65a3fabf063064395b8846422ce964d59dda2ce7/crates/blockifier/src/execution/entry_point.rs#L73)):*
> ```XML
>         let mut decrement_when_dropped = RecursionDepthGuard::new(
>             context.current_recursion_depth.clone(),
>             context.block_context.versioned_constants.max_recursion_depth,
> ```

In your feature branch (that contains the whole feature): how many `a.b.c.d`s do you have? Should we consider encapsulating some of them with getters?



_Code quote:_
```XML
context.block_context.versioned_constants
```
___
*[`crates/native_blockifier/src/py_block_executor.rs` line 291 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Noq2AEbCSzEoRD9py6j:-Noq2AEbCSzEoRD9py6k:bx0oeze) ([raw file](https://github.com/starkware-libs/blockifier/blob/65a3fabf063064395b8846422ce964d59dda2ce7/crates/native_blockifier/src/py_block_executor.rs#L291)):*
> ```XML
>     general_config: &PyGeneralConfig,
>     block_info: PyBlockInfo,
> ) -> NativeBlockifierResult<BlockContext> {
> ```

Why so many `clone()`s? :(
___
*[`crates/native_blockifier/src/py_validator.rs` line 25 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Noq2HST-0eBFXmKUrzJ:-Noq2HST-0eBFXmKUrzK:b676hhe) ([raw file](https://github.com/starkware-libs/blockifier/blob/016c16557b558e8cba0fc008c9e602beb60d4e42/crates/native_blockifier/src/py_validator.rs#L25)):*
> ```XML
> #[pyclass]
> pub struct PyValidator {
>     pub general_config: PyGeneralConfig,
> ```

How come this is not needed?



_Code quote:_
```XML
pub general_config: PyGeneralConfig,
```

elintul (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: all files reviewed, 9 unresolved discussions (waiting on @giladchase and @noaov1)

___
*[`crates/blockifier/resources/versioned_constants.json` line 16 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NoqAmUn8X3jkO0cpRQf:-NoqAmUn8X3jkO0cpRQg:b-igmh8d) ([raw file](https://github.com/starkware-libs/blockifier/blob/4ad369687499397fbaa893dd68801ccf668f3a3d/crates/blockifier/resources/versioned_constants.json#L16)):*
> ```JSON
>     "invoke_tx_max_n_steps": 3000000,
>     "validate_max_n_steps": 1000000,
>     "max_recursion_depth": 50
> ```

Not sure we want it here; it's not really versioned, but should stay configurable.



_Code quote:_
```JSON
"max_recursion_depth": 50
```

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: 21 of 24 files reviewed, 6 unresolved discussions (waiting on @elintul and @noaov1)

___
*[`crates/blockifier/src/versioned_constants.rs` line 19 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NovTe8wBTlH_vU58LYg:-NovTe8wBTlH_vU58LYh:b-xom2ur) ([raw file](https://github.com/starkware-libs/blockifier/blob/81ade96d44d16c13a2a100d8160cbb616a102d6e/crates/blockifier/src/versioned_constants.rs#L19)):*
> ```XML
> pub struct VersionedConstants {
>     // Fee related.
>     pub vm_resource_fee_cost: Arc<HashMap<String, f64>>,
> ```

Should this be `pub vm_resource_fee_cost: ResourceCosts` (where `ResourceCosts` contains the 8 values in the json)?
I think it might be better, because the HashMap implies that this has a dynamic structure, even though (IIUC) the keys in this map change very rarely and we won't have too many of them.



_Code quote:_
```XML
    pub vm_resource_fee_cost: Arc<HashMap<String, f64>>,
```

elintul (changes requested):
> Reviewed 2 of 5 files at r4, 1 of 6 files at r5.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: all files reviewed, 9 unresolved discussions (waiting on @giladchase and @noaov1)

___
*[`crates/blockifier/src/versioned_constants.rs` line 19 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NovTe8wBTlH_vU58LYg:-Np-DLs83fR4m4wfzxzB:bb77ufg) ([raw file](https://github.com/starkware-libs/blockifier/blob/81ade96d44d16c13a2a100d8160cbb616a102d6e/crates/blockifier/src/versioned_constants.rs#L19)):*

Should this be `pub vm_resource_fee_cost: ResourceCosts` (where `ResourceCosts` contains the 8 values in the json)?
I think it might be better, because the HashMap implies that this has a dynamic structure, even though (IIUC) the keys in this map change very rarely and we won't have too many of them.
</blockquote></details>

Yes.
___
*[`crates/blockifier/src/versioned_constants.rs` line 44 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Np-DOKf5-rjqmPr2q3e:-Np-DOKf5-rjqmPr2q3f:b2xiez8) ([raw file](https://github.com/starkware-libs/blockifier/blob/81ade96d44d16c13a2a100d8160cbb616a102d6e/crates/blockifier/src/versioned_constants.rs#L44)):*
> ```XML
> 
> #[derive(Debug, Error)]
> pub enum VersionedConstantsError {
> ```

Alphabetize, please.



_Code quote:_
```XML
VersionedConstantsError
```
___
*[`crates/native_blockifier/src/py_block_executor.rs` line 291 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Noq2AEbCSzEoRD9py6j:-Np-D2I90RzLZp2QoZGb:b-ucfknh) ([raw file](https://github.com/starkware-libs/blockifier/blob/65a3fabf063064395b8846422ce964d59dda2ce7/crates/native_blockifier/src/py_block_executor.rs#L291)):*

First one (starknet_os_config): is because `ChainInfo` wants to own its fields; it's fields are small enough to not warrent an Arc, and adding a reference to this struct will require a lifetime, which, for a context struct, will pollute a lot of the methods with lifetimes.

Second one: (`chain_info: chain_info.clone()`) that's a bug, unnecessary clone, removed. Thanks for spotting :pray: 

Thid one (versioned_constants): BlockContext wants to own this instance. 
We can make this Arc: I don't think this is warrent, because the instance is almost a `Copy`; it has 3 `Copy` fields and one `Arc` field, which is cheapish to clone (but not `Copy`-cheap).
I moved the instance to the py_block_executor because of the override support i added for max_recursion, the copy is still necessary though, for the same reason.
</blockquote></details>

Thanks!
___
*[`crates/native_blockifier/src/py_block_executor.rs` line 51 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Np-DE5C0SCgmgmlA-Aq:-Np-DE5C0SCgmgmlA-Ar:b-m1keq0) ([raw file](https://github.com/starkware-libs/blockifier/blob/81ade96d44d16c13a2a100d8160cbb616a102d6e/crates/native_blockifier/src/py_block_executor.rs#L51)):*
> ```XML
>             max_recursion_depth,
>             ..VersionedConstants::latest_constants().clone()
>         };
> ```

Repeated three times, consider defining a c-tor.



_Code quote:_
```XML
        let versioned_constants = VersionedConstants {
            max_recursion_depth,
            ..VersionedConstants::latest_constants().clone()
        };
```
___
*[`crates/native_blockifier/src/py_block_executor.rs` line 60 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Nov50v40eIraW04WFBU:-Nov50v40eIraW04WFBV:undefined) ([raw file](https://github.com/starkware-libs/blockifier/blob/81ade96d44d16c13a2a100d8160cbb616a102d6e/crates/native_blockifier/src/py_block_executor.rs#L60)):*
> ```XML
>             storage: Box::new(storage),
>             global_contract_cache: GlobalContractCache::default(),
>         }
> ```





_Suggestion:_
```XML
        log::debug!("Initializing Block Executor...");
        let storage =
            PapyrusStorage::new(target_storage_config).expect("Failed to initialize storage");
        let versioned_constants = VersionedConstants {
            max_recursion_depth,
            ..VersionedConstants::latest_constants().clone()
        };
        log::debug!("Initialized Block Executor.");
        
        Self {
            general_config,
            versioned_constants,
            tx_executor: None,
            storage: Box::new(storage),
            global_contract_cache: GlobalContractCache::default(),
        }
```
___
*[`crates/native_blockifier/src/py_block_executor.rs` line 204 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Np-Ck5M9rwBmojgNhiQ:-Np-Ck5M9rwBmojgNhiR:bhdgk2k) ([raw file](https://github.com/starkware-libs/blockifier/blob/a743d49cf716eefb5178acc4c047b6de7028fb7c/crates/native_blockifier/src/py_block_executor.rs#L204)):*
> ```XML
>             )),
>             general_config,
>             max_recursion_depth: 50,
> ```

So this appears inside the constants and passed from Python configurably?



_Code quote:_
```XML
max_recursion_depth: 50,
```

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: all files reviewed, 4 unresolved discussions (waiting on @elintul and @noaov1)

___
*[`crates/blockifier/resources/versioned_constants.json` line 15 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Noq2fmm0IAgCqz6QWIh:-Np-kg2BDvg73LE8FZ8W:bs6xklw) ([raw file](https://github.com/starkware-libs/blockifier/blob/65a3fabf063064395b8846422ce964d59dda2ce7/crates/blockifier/resources/versioned_constants.json#L15)):*

Where did they come from?
</blockquote></details>

Python, searched the configs/ymls/presets and they all matched to these values (excpe
___
*[`crates/blockifier/src/versioned_constants.rs` line 19 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NovTe8wBTlH_vU58LYg:-Np-mLPKAoKdr5CDLpwk:b-69wb80) ([raw file](https://github.com/starkware-libs/blockifier/blob/81ade96d44d16c13a2a100d8160cbb616a102d6e/crates/blockifier/src/versioned_constants.rs#L19)):*

Yes.
</blockquote></details>

Ahh poop looks like this isn't trivial, we have a bunch of logic that already assumes this is a hash. 
I added a TODO, that OK?
___
*[`crates/blockifier/src/versioned_constants.rs` line 44 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Np-DOKf5-rjqmPr2q3e:-Np-mo5i0d2ZHeRpxwii:b-in7lqn) ([raw file](https://github.com/starkware-libs/blockifier/blob/81ade96d44d16c13a2a100d8160cbb616a102d6e/crates/blockifier/src/versioned_constants.rs#L44)):*

Alphabetize, please.
</blockquote></details>

Right, srry
___
*[`crates/blockifier/src/execution/entry_point.rs` line 73 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Noq196875y4etQEX2WT:-Np-nA_k7W4Uy34klJq_:bibyfkt) ([raw file](https://github.com/starkware-libs/blockifier/blob/65a3fabf063064395b8846422ce964d59dda2ce7/crates/blockifier/src/execution/entry_point.rs#L73)):*

Can you please add this as a task on Monday so we won't forget? Not urgent for `0.13.1`, obviously.
</blockquote></details>

Done https://starkware.monday.com/boards/2735879357/pulses/5930145121
___
*[`crates/native_blockifier/src/py_block_executor.rs` line 51 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Np-DE5C0SCgmgmlA-Aq:-Np-o65z-2XM4s3qicFC:bpgd8qt) ([raw file](https://github.com/starkware-libs/blockifier/blob/81ade96d44d16c13a2a100d8160cbb616a102d6e/crates/native_blockifier/src/py_block_executor.rs#L51)):*

Repeated three times, consider defining a c-tor.
</blockquote></details>

Done (i only found 2 though :scream: )
___
*[`crates/native_blockifier/src/py_block_executor.rs` line 204 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Np-Ck5M9rwBmojgNhiQ:-Np-oDwq7Zv_6Ato0Rts:b2wi487) ([raw file](https://github.com/starkware-libs/blockifier/blob/a743d49cf716eefb5178acc4c047b6de7028fb7c/crates/native_blockifier/src/py_block_executor.rs#L204)):*

So this appears inside the constants and passed from Python configurably?
</blockquote></details>

Yep, basically the value we get from python completly overrides the value in the json.
We deserialize the json and immediately replace the value with what we got from python.
This is a strictly native_blockifier thing of course, normal blockifier isn't affected.

elintul (changes requested):
> Reviewed 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: all files reviewed, 5 unresolved discussions (waiting on @giladchase and @noaov1)

___
*[`crates/blockifier/resources/versioned_constants.json` line 15 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Noq2fmm0IAgCqz6QWIh:-Np-pVkS9pMPHXY1hBKS:b-v1p2el) ([raw file](https://github.com/starkware-libs/blockifier/blob/65a3fabf063064395b8846422ce964d59dda2ce7/crates/blockifier/resources/versioned_constants.json#L15)):*

Python, searched the configs/ymls/presets and they all matched to these values (excpe
</blockquote></details>

Are these mainnet values?
___
*[`crates/blockifier/src/versioned_constants.rs` line 19 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NovTe8wBTlH_vU58LYg:-Np-pbr837HyQgK1DG4I:bod2t8g) ([raw file](https://github.com/starkware-libs/blockifier/blob/81ade96d44d16c13a2a100d8160cbb616a102d6e/crates/blockifier/src/versioned_constants.rs#L19)):*

Ahh poop looks like this isn't trivial, we have a bunch of logic that already assumes this is a hash. 
I added a TODO, that OK?
</blockquote></details>

Yup. What logic assumes it's a hash, for example?
___
*[`crates/native_blockifier/src/py_block_executor.rs` line 51 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Np-DE5C0SCgmgmlA-Aq:-Np-ps4S2-3uiK1gBXMu:bmkbhv7) ([raw file](https://github.com/starkware-libs/blockifier/blob/81ade96d44d16c13a2a100d8160cbb616a102d6e/crates/native_blockifier/src/py_block_executor.rs#L51)):*

Done (i only found 2 though :scream: )
</blockquote></details>

Chrome lies?
___
*[`crates/native_blockifier/src/py_block_executor.rs` line 204 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Np-Ck5M9rwBmojgNhiQ:-Np-puV83xKLCiqRT-Jo:b-unyo8x) ([raw file](https://github.com/starkware-libs/blockifier/blob/a743d49cf716eefb5178acc4c047b6de7028fb7c/crates/native_blockifier/src/py_block_executor.rs#L204)):*

Yep, basically the value we get from python completly overrides the value in the json.
We deserialize the json and immediately replace the value with what we got from python.
This is a strictly native_blockifier thing of course, normal blockifier isn't affected.
</blockquote></details>

Isn't this a bit confusing?
___
*[`crates/native_blockifier/src/py_utils.rs` line 103 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Np-pop8CLdsaMkxVfrV:-Np-pop8CLdsaMkxVfrW:b-1iyfyd) ([raw file](https://github.com/starkware-libs/blockifier/blob/2679032881ac115c8951bb05e522d11511e393e1/crates/native_blockifier/src/py_utils.rs#L103)):*
> ```XML
> }
> 
> pub fn versioned_constants_with_overrides(max_recursion_depth: usize) -> VersionedConstants {
> ```

Why not a c-tor?



_Code quote:_
```XML
versioned_constants_with_overrides
```

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: all files reviewed, 5 unresolved discussions (waiting on @elintul and @noaov1)

___
*[`crates/blockifier/resources/versioned_constants.json` line 15 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Noq2fmm0IAgCqz6QWIh:-Np0BuaL0s3VbxnAnOAp:b-c8ppl0) ([raw file](https://github.com/starkware-libs/blockifier/blob/65a3fabf063064395b8846422ce964d59dda2ce7/crates/blockifier/resources/versioned_constants.json#L15)):*

Are these mainnet values?
</blockquote></details>

Yep, and as far as i can find also the values for other chains, but i didn't do a thorough check
___
*[`crates/blockifier/src/versioned_constants.rs` line 19 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NovTe8wBTlH_vU58LYg:-Np0CNz8EmExj0-rnBZZ:b4urpes) ([raw file](https://github.com/starkware-libs/blockifier/blob/81ade96d44d16c13a2a100d8160cbb616a102d6e/crates/blockifier/src/versioned_constants.rs#L19)):*

Yup. What logic assumes it's a hash, for example?
</blockquote></details>

[This function](https://github.com/starkware-libs/blockifier/blob/0c2e2ffc1183b469998286569412076128568727/crates/blockifier/src/fee/fee_utils.rs#L40)
heavily uses the hashiness.
Probably can get around this in a nice way with a struct, but it'll take some code and testing to make sure it's OK, maybe out of scope for this one?
___
*[`crates/native_blockifier/src/py_block_executor.rs` line 51 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Np-DE5C0SCgmgmlA-Aq:-Np0CcoeBnc1DEcJAP03:bbotw8i) ([raw file](https://github.com/starkware-libs/blockifier/blob/81ade96d44d16c13a2a100d8160cbb616a102d6e/crates/native_blockifier/src/py_block_executor.rs#L51)):*

Chrome lies?
</blockquote></details>

DOS prevention?
___
*[`crates/native_blockifier/src/py_block_executor.rs` line 204 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Np-Ck5M9rwBmojgNhiQ:-Np0D5Qr33LLN4dcWQKB:bqjttbf) ([raw file](https://github.com/starkware-libs/blockifier/blob/a743d49cf716eefb5178acc4c047b6de7028fb7c/crates/native_blockifier/src/py_block_executor.rs#L204)):*

Isn't this a bit confusing?
</blockquote></details>

A bit ya, it was just the simplest thing i could do.
A better way of doing this (also answers the overide constructor thread below is to make an override struct which can consume itself and create a VersionedConstants instance (something similar to a builder pattern).

But since there was only one override here (+ validate_max_n_steps later) i thought i might not be worth the overhead, but if you think it is i'll add it of course :)
___
*[`crates/native_blockifier/src/py_utils.rs` line 103 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Np-pop8CLdsaMkxVfrV:-Np0Coj4DguIztPYhv7t:b-fezc06) ([raw file](https://github.com/starkware-libs/blockifier/blob/2679032881ac115c8951bb05e522d11511e393e1/crates/native_blockifier/src/py_utils.rs#L103)):*

Why not a c-tor?
</blockquote></details>

I thought about it, but since this is very much a Python thing, i thought it might fit better in native_blockifier.

In other words, we don't have any specific max_recursion override in the blockifier, so adding a constructor there seems out of place.

elintul (commented):
> Reviewed 13 of 13 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: all files reviewed, 1 unresolved discussion (waiting on @noaov1)

noaov1 (changes requested):
> Reviewed 6 of 24 files at r1, 2 of 3 files at r2, 1 of 5 files at r4, 1 of 6 files at r5, 2 of 4 files at r6, 13 of 13 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: all files reviewed, 5 unresolved discussions (waiting on @elintul and @giladchase)

___
*[`crates/blockifier/BUILD` line 7 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NpOvCTY8Jze-ZnzT16Q:-NpOvCTZByU86N94uInj:b-cxk7zb) ([raw file](https://github.com/starkware-libs/blockifier/blob/2e0b21915a9ac7ba8155f40267790f5f25cde7f4/crates/blockifier/BUILD#L7)):*
> ```Starlark
>     name="blockifier",
>     srcs=glob(["src/**/*.rs"]),
>     data=glob(["resources/**/*"]),
> ```

Can you please explain this change? :)



_Code quote:_
```Starlark
    data=glob(["resources/**/*"]),
```
___
*[`crates/blockifier/resources/versioned_constants.json` line 15 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Noq2fmm0IAgCqz6QWIh:-NpOzeHP3MDaOczwtuc9:b1i82zr) ([raw file](https://github.com/starkware-libs/blockifier/blob/65a3fabf063064395b8846422ce964d59dda2ce7/crates/blockifier/resources/versioned_constants.json#L15)):*

Yep, and as far as i can find also the values for other chains, but i didn't do a thorough check
</blockquote></details>

Why do [here](https://github.com/starkware-industries/starkware/blob/b3df9fca8b5cb0d17ad7c192694946b6de096125/src/starkware/starknet/definitions/general_config.yml#L17) we set `invoke_tx_max_n_steps: 1000000`?
___
*[`crates/blockifier/resources/versioned_constants.json` line 1 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NpP4pAe2Xl81BM5Qsxt:-NpP4pAe2Xl81BM5Qsxu:bkvo42m) ([raw file](https://github.com/starkware-libs/blockifier/blob/2e0b21915a9ac7ba8155f40267790f5f25cde7f4/crates/blockifier/resources/versioned_constants.json#L1)):*
> ```JSON
> {
> ```

Q- Why is this file located here and not under the `versioned_constants` module?
___
*[`crates/blockifier/src/execution/entry_point.rs` line 225 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NpP8Le14tlnnflgCZNO:-NpP8Le14tlnnflgCZNP:bv5ub8n) ([raw file](https://github.com/starkware-libs/blockifier/blob/2e0b21915a9ac7ba8155f40267790f5f25cde7f4/crates/blockifier/src/execution/entry_point.rs#L225)):*
> ```XML
>             ExecutionMode::Execute => min(
>                 constants.invoke_tx_max_n_steps,
>                 default_versioned_constants.invoke_tx_max_n_steps,
> ```

Why is the `min` needed? Isn't it the same value? Will we allow to override this value?
Wouldn't that be problematic for a full node re-executing transactions from older versions? 



_Code quote:_
```XML
            ExecutionMode::Execute => min(
                constants.invoke_tx_max_n_steps,
                default_versioned_constants.invoke_tx_max_n_steps,
```
___
*[`crates/native_blockifier/src/py_validator.rs` line 48 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NpP88DA1SyNOhu3fWFC:-NpP88DA1SyNOhu3fWFD:b-361n9t) ([raw file](https://github.com/starkware-libs/blockifier/blob/2e0b21915a9ac7ba8155f40267790f5f25cde7f4/crates/native_blockifier/src/py_validator.rs#L48)):*
> ```XML
>             &general_config,
>             // TODO(Gilad): add max_validate_n_steps override argument and override here.
>             &versioned_constants,
> ```

Can you please explain why? :)



_Code quote:_
```XML
            // TODO(Gilad): add max_validate_n_steps override argument and override here.
            &versioned_constants,
```

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: 23 of 25 files reviewed, 5 unresolved discussions (waiting on @elintul and @noaov1)

___
*[`Cargo.toml` line 55 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NpPh5Ff2EzV8ZrOmUl8:-NpPh5Ff2EzV8ZrOmUl9:bdku9c) ([raw file](https://github.com/starkware-libs/blockifier/blob/2e0b21915a9ac7ba8155f40267790f5f25cde7f4/Cargo.toml#L55)):*
> ```TOML
> tempfile = "3.7.0"
> test-case = "2.2.2"
> thiserror = "1.0.37"
> ```

alphabetized + added once_cell for `Lazy`



_Code quote:_
```TOML
once_cell = "1.19.0"
papyrus_storage = "0.3.0-dev.0"
phf = { version = "0.11", features = ["macros"] }
pretty_assertions = "1.2.1"
pyo3 = "0.19.1"
pyo3-log = "0.8.1"
rstest = "0.17.0"
serde = "1.0.184"
serde_json = "1.0.81"
sha3 = "0.10.6"
starknet-crypto = "0.5.1"
starknet_api = "0.7.0-dev.0"
strum = "0.24.1"
strum_macros = "0.24.3"
tempfile = "3.7.0"
test-case = "2.2.2"
thiserror = "1.0.37"
```
___
*[`crates/blockifier/resources/versioned_constants.json` line 15 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Noq2fmm0IAgCqz6QWIh:-NpPp5I43SMtrqhXIuZn:b-aycx7j) ([raw file](https://github.com/starkware-libs/blockifier/blob/65a3fabf063064395b8846422ce964d59dda2ce7/crates/blockifier/resources/versioned_constants.json#L15)):*

Why do [here](https://github.com/starkware-industries/starkware/blob/b3df9fca8b5cb0d17ad7c192694946b6de096125/src/starkware/starknet/definitions/general_config.yml#L17) we set `invoke_tx_max_n_steps: 1000000`?
</blockquote></details>

Not sure really, the values might vary between chains, and some might be stale and always overwritten.
I used 3000000 because that seemed to be the [default value](https://github.com/starkware-industries/starkware/blob/07e3265bd768d004b48ba7120aab433a112e13fc/src/starkware/starknet/definitions/general_config.py#L172) for general_config, and the expected ymls from the pyconf test also seemed to agree.

There's a task to go over all chains and compare values, which i'm doing concurrently to this, which was agreed to be mainnet values.

I think mainnet uses the default value (3000000) because its preset doesn't seem to `Replace` the default value above.
I can't really check for sure cause i don't have mainnet access.
___
*[`crates/blockifier/resources/versioned_constants.json` line 1 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NpP4pAe2Xl81BM5Qsxt:-NpPo-hD5i_wH5LFS-3T:bhld876) ([raw file](https://github.com/starkware-libs/blockifier/blob/2e0b21915a9ac7ba8155f40267790f5f25cde7f4/crates/blockifier/resources/versioned_constants.json#L1)):*

Q- Why is this file located here and not under the `versioned_constants` module?
</blockquote></details>

The `resources` dir is typically where ppl put non-code stuff, that is intended to be consumed during execution.
Putting it anywhere inside `src` could be considered a bit inappropriate, since it isn't source code, it's static.

Moreover, since this file should be easily accessible for users, and even allow them to store more similar files in this directory in addition to this one for quick switching between constants, a central location like a dedicated dir at the root of the crate is a spot for this.
The alternative would be to send the users to dig into our code, which is less convenient (especially for product ppl who just want to know prices of stuff and don't want to dig around code).
___
*[`crates/blockifier/src/versioned_constants.rs` line 13 at r4](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NouZpo37AjmivPjpdFs:-NpPqt7b68C_jGlhmw1b:b7i3wbz) ([raw file](https://github.com/starkware-libs/blockifier/blob/1f59fa544b456442854ff43bc6c33bdab222632b/crates/blockifier/src/versioned_constants.rs#L13)):*
> ```XML
> static DEFAULT_CONSTANTS: Lazy<VersionedConstants> = Lazy::new(|| {
>     serde_json::from_str(DEFAULT_CONSTANTS_JSON)
>         .expect("Versioned constants json file is malformed")
> ```

Done, srry about the delay, i fixed it and must have lost it in a rebase.
___
*[`crates/blockifier/src/execution/entry_point.rs` line 225 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NpP8Le14tlnnflgCZNO:-NpPqzDZ-Dw3CQzB4s0N:b-gvzy88) ([raw file](https://github.com/starkware-libs/blockifier/blob/2e0b21915a9ac7ba8155f40267790f5f25cde7f4/crates/blockifier/src/execution/entry_point.rs#L225)):*
>Why is the `min` needed? Isn't it the same value?

For us it will be the same value, ya.
But for a node, if they load up a different json file this means "take the minimum between the `invoke_tx_mx_n_steps` from the json file you loaded and the one that came with json file that was shipped with this blockifier version (aka "default versioned_constants" here).

This can happen if they want to preload new gas costs without bumping a blockifier version (which might not work properly if the blockifier version with the new gas costs had other logic changes but this feature is a best-effort kind of thing).

>Will we allow to override this value?  
>Wouldn't that be problematic for a full node re-executing transactions from older versions?

We won't override it interanally, for the exact reason you mentioned.
This isn't the case for max_validate n_steps, since that is only relevant for the gateway, and the nodes have nothing to do with that bit.
___
*[`crates/native_blockifier/src/py_validator.rs` line 48 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NpP88DA1SyNOhu3fWFC:-NpPs77B6SZT73A8yFmB:bngrr31) ([raw file](https://github.com/starkware-libs/blockifier/blob/2e0b21915a9ac7ba8155f40267790f5f25cde7f4/crates/native_blockifier/src/py_validator.rs#L48)):*

Can you please explain why? :)
</blockquote></details>

Basically we want to retain override capability for any consts that won't mess up the nodes (for example max_invoke n steps, like you mentioned in one of the comments above).
Changing max validate on the fly won't mess up nodes (at most, they'll start only getting txs with simple validates, but that's about it).
___
*[`crates/blockifier/BUILD` line 7 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NpOvCTY8Jze-ZnzT16Q:-NpPhDAOE2MXWYmk8ZPq:b-tp07kw) ([raw file](https://github.com/starkware-libs/blockifier/blob/2e0b21915a9ac7ba8155f40267790f5f25cde7f4/crates/blockifier/BUILD#L7)):*

Can you please explain this change? :)
</blockquote></details>

Hmm that's not needed anymore actually :scream: 

Removed.

Remained from a previous implementation, when i was accessing the json file via `native_blockifier`.

This change makes these files accessible to bazel.
However, since now we're downloading the file directly from blockifier's github page, this isn't necessary.

elintul (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: all files reviewed, 4 unresolved discussions (waiting on @noaov1)

elintul (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-:-NpQKFZy8pFkjj7EF-PC:bnfp4nl)* status: all files reviewed, 4 unresolved discussions (waiting on @noaov1)

noaov1 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: all files reviewed, 3 unresolved discussions (waiting on @giladchase)

___
*[`crates/blockifier/resources/versioned_constants.json` line 1 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NpP4pAe2Xl81BM5Qsxt:-NpQGibL5XkUf6ofjAcK:b-jhyyta) ([raw file](https://github.com/starkware-libs/blockifier/blob/2e0b21915a9ac7ba8155f40267790f5f25cde7f4/crates/blockifier/resources/versioned_constants.json#L1)):*

The `resources` dir is typically where ppl put non-code stuff, that is intended to be consumed during execution.
Putting it anywhere inside `src` could be considered a bit inappropriate, since it isn't source code, it's static.

Moreover, since this file should be easily accessible for users, and even allow them to store more similar files in this directory in addition to this one for quick switching between constants, a central location like a dedicated dir at the root of the crate is a spot for this.
The alternative would be to send the users to dig into our code, which is less convenient (especially for product ppl who just want to know prices of stuff and don't want to dig around code).
</blockquote></details>

Got it. Thanks!
___
*[`crates/blockifier/src/execution/entry_point.rs` line 225 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NpP8Le14tlnnflgCZNO:-NpQIkVJCv71WVyTuPzs:b-x18zi2) ([raw file](https://github.com/starkware-libs/blockifier/blob/2e0b21915a9ac7ba8155f40267790f5f25cde7f4/crates/blockifier/src/execution/entry_point.rs#L225)):*
> But for a node, if they load up a different json file this means "take the minimum between the `invoke_tx_mx_n_steps` from the json file you loaded and the one that came with json file that was shipped with this blockifier version (aka "default versioned\_constants" here).

Why do we want this behaviour?
Isn't it enough to use only `constants.invoke_tx_max_n_steps`?
___
*[`crates/native_blockifier/src/py_validator.rs` line 48 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NpP88DA1SyNOhu3fWFC:-NpQJHGp6kc3XXDg2UcB:bb3643f) ([raw file](https://github.com/starkware-libs/blockifier/blob/2e0b21915a9ac7ba8155f40267790f5f25cde7f4/crates/native_blockifier/src/py_validator.rs#L48)):*

Basically we want to retain override capability for any consts that won't mess up the nodes (for example max_invoke n steps, like you mentioned in one of the comments above).
Changing max validate on the fly won't mess up nodes (at most, they'll start only getting txs with simple validates, but that's about it).
</blockquote></details>

I don't think I fully understand :)
Isn't changing the `max_validate_n_steps` can affect the nodes when they re-execute transactions? (assume they are using the values from the file while we override it in the original execution)

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: all files reviewed, 2 unresolved discussions (waiting on @noaov1)

___
*[`crates/blockifier/src/execution/entry_point.rs` line 225 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NpP8Le14tlnnflgCZNO:-NpQbSkP0-VjoQVds7S2:b-owyi8f) ([raw file](https://github.com/starkware-libs/blockifier/blob/2e0b21915a9ac7ba8155f40267790f5f25cde7f4/crates/blockifier/src/execution/entry_point.rs#L225)):*

> But for a node, if they load up a different json file this means "take the minimum between the `invoke_tx_mx_n_steps` from the json file you loaded and the one that came with json file that was shipped with this blockifier version (aka "default versioned\_constants" here).

Why do we want this behaviour?
Isn't it enough to use only `constants.invoke_tx_max_n_steps`?
</blockquote></details>

Good question, actually now that i think about it maybe constants::MAX_VALIDATE_STEPS_PER_TX was intended as global maximum :thinking: 
I'll ask Dori tomorrow and replace this const, i think it really was a global maximum.

Eitherway, good catch :eagle:

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: 23 of 25 files reviewed, 2 unresolved discussions (waiting on @elintul and @noaov1)

___
*[`crates/native_blockifier/src/py_validator.rs` line 48 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NpP88DA1SyNOhu3fWFC:-NpTF60Q3Gg5vT8cfZCs:b-85m878) ([raw file](https://github.com/starkware-libs/blockifier/blob/2e0b21915a9ac7ba8155f40267790f5f25cde7f4/crates/native_blockifier/src/py_validator.rs#L48)):*

I don't think I fully understand :)
Isn't changing the `max_validate_n_steps` can affect the nodes when they re-execute transactions? (assume they are using the values from the file while we override it in the original execution)
</blockquote></details>

This is added up the stack, i'll add a TODO there to check it out, plz unblock for now? :grimacing:

noaov1 (dismissed):
> Reviewed 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @giladchase)

elintul (changes requested):
> Reviewed 1 of 2 files at r9, 16 of 16 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: all files reviewed, 1 unresolved discussion (waiting on @giladchase)

___
*[`Cargo.toml` line 16 at r10](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NpU8Xsw7UTs8ky5E--O:-NpU8Xsw7UTs8ky5E--P:b-ye04is) ([raw file](https://github.com/starkware-libs/blockifier/blob/ef3526a6d9039859b1260dedfecaa90305d0c655/Cargo.toml#L16)):*
> ```TOML
> 
> [workspace.dependencies]
> anyhow = "1.0.0"
> ```

Why is it back?



_Code quote:_
```TOML
anyhow = "1.0.0"
```

elintul (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @giladchase)

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @giladchase)

___
*[`Cargo.toml` line 16 at r10](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NpU8Xsw7UTs8ky5E--O:-NpUA7uz-5Mx3QkpaXQA:b-2v85yx) ([raw file](https://github.com/starkware-libs/blockifier/blob/ef3526a6d9039859b1260dedfecaa90305d0c655/Cargo.toml#L16)):*

Why is it back?
</blockquote></details>

Rebased, someone added it main.
I think the reason had to do with the VM using it as a return type :(

Anyway not related to my work.

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @giladchase)

___
*[`crates/blockifier/resources/versioned_constants.json` line 15 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-Noq2fmm0IAgCqz6QWIh:-NpUAL8u4oYQBblZ21rX:b-pvm0jp) ([raw file](https://github.com/starkware-libs/blockifier/blob/65a3fabf063064395b8846422ce964d59dda2ce7/crates/blockifier/resources/versioned_constants.json#L15)):*

You can run the deployment command with the mainnet preset (or any env., for this matter) with `--dry_run` flag and TAL at the created artifacts.
</blockquote></details>

Noa checked, the `1000000` is just for testing.
Elin: i'll run the command you said after we merge the stack? worst case i'll just have to change the json values again.

elintul (dismissed):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @giladchase)

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: 24 of 25 files reviewed, all discussions resolved (waiting on @elintul)

___
*[`crates/blockifier/BUILD` line 7 at r11](https://reviewable.io/reviews/starkware-libs/blockifier/1348#-NpUXb77BFPB880_qyWl:-NpUXb78E4PVwKh2Bfvt:bw4egzn) ([raw file](https://github.com/starkware-libs/blockifier/blob/04b49fa49f183ba30b422a56a2c1830765c56d36/crates/blockifier/BUILD#L7)):*
> ```Starlark
>     name="blockifier",
>     srcs=glob(["src/**/*.rs"]),
>     data=glob(["resources/**/*"]),
> ```

OK this ended up being important for bazel, without this it won't have access to these files, they have to be whitelisted, which is basically what this does



_Code quote:_
```Starlark
    data=glob(["resources/**/*"]),
```

elintul (approved):
> Reviewed 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1348)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @giladchase)

PR #1646, opened 2024-03-07
refactor: added events cost to StarknetResources

nimrod-starkware:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1646)

### DIFF
--- a/crates/blockifier/src/fee/actual_cost.rs
+++ b/crates/blockifier/src/fee/actual_cost.rs
@@ -9,7 +9,6 @@ use crate::abi::constants as abi_constants;
 use crate::context::TransactionContext;
 use crate::execution::call_info::CallInfo;
 use crate::execution::contract_class::ClassInfo;
-use crate::fee::gas_usage::get_tx_events_gas_cost;
 use crate::state::cached_state::{CachedState, StateChanges, StateChangesCount};
 use crate::state::state_api::{StateReader, StateResult};
 use crate::transaction::objects::{
@@ -17,7 +16,6 @@ use crate::transaction::objects::{
 };
 use crate::transaction::transaction_types::TransactionType;
 use crate::transaction::transaction_utils::calculate_tx_resources;
-use crate::versioned_constants::VersionedConstants;
 
 #[cfg(test)]
 #[path = "actual_cost_test.rs"]
@@ -165,20 +163,12 @@ impl<'a> ActualCostBuilder<'a> {
         let non_optional_call_infos =
             self.validate_call_info.into_iter().chain(self.execute_call_info);
 
-        self.starknet_resources.set_messages_resources(non_optional_call_infos.clone())?;
-        // Gas usage for SHARP costs and Starknet L1-L2 messages. Includes gas usage for data
-        // availability.
-        let gas_usage_vector = Self::calculate_tx_gas_usage_vector(
-            &self.tx_context.block_context.versioned_constants,
-            non_optional_call_infos,
-            &self.starknet_resources,
-            use_kzg_da,
-        )?;
+        // Set the events and messages resources from the transaction's call infos.
+        self.starknet_resources.set_events_and_messages_resources(non_optional_call_infos)?;
 
         let mut actual_resources = calculate_tx_resources(
             &self.tx_context.block_context.versioned_constants,
             execution_resources,
-            gas_usage_vector,
             self.tx_type,
             &self.starknet_resources,
             use_kzg_da,
@@ -202,20 +192,5 @@ impl<'a> ActualCostBuilder<'a> {
         };
 
         Ok((ActualCost { actual_fee, da_gas, actual_resources }, bouncer_resources))
-    }
-
-    /// Returns the gas usage of a transaction, specifically:
-    /// * L1 gas, used by Starknet's state update and the Verifier, e.g., a message from L2 to L1 is
-    ///   followed by a storage write operation on L1.
-    /// * L1 data gas, for publishing data availability.
-    /// * L2 resources cost, e.g., for storing transaction calldata.
-    fn calculate_tx_gas_usage_vector(
-        versioned_constants: &VersionedConstants,
-        call_infos: impl Iterator<Item = &'a CallInfo> + Clone,
-        starknet_resources: &StarknetResources,
-        use_kzg_da: bool,
-    ) -> TransactionExecutionResult<GasVector> {
-        Ok(starknet_resources.to_gas_vector(versioned_constants, use_kzg_da)
-            + get_tx_events_gas_cost(call_infos, versioned_constants))
     }
 }
--- a/crates/blockifier/src/fee/actual_cost_test.rs
+++ b/crates/blockifier/src/fee/actual_cost_test.rs
@@ -6,7 +6,6 @@ use starknet_api::transaction::{Fee, L2ToL1Payload, TransactionVersion};
 
 use crate::context::BlockContext;
 use crate::execution::call_info::{CallExecution, CallInfo, MessageToL1, OrderedL2ToL1Message};
-use crate::fee::actual_cost::ActualCostBuilder;
 use crate::fee::eth_gas_constants;
 use crate::fee::gas_usage::{
     get_consumed_message_to_l2_emissions_cost, get_log_message_to_l1_emissions_cost,
@@ -45,13 +44,9 @@ fn versioned_constants() -> &'static VersionedConstants {
 fn test_calculate_tx_gas_usage_basic<'a>(#[values(false, true)] use_kzg_da: bool) {
     // An empty transaction (a theoretical case for sanity check).
     let versioned_constants = VersionedConstants::default();
-    let empty_tx_gas_usage_vector = ActualCostBuilder::calculate_tx_gas_usage_vector(
-        &versioned_constants,
-        std::iter::empty(),
-        &StarknetResources::default(),
-        use_kzg_da,
-    )
-    .unwrap();
+    let empty_tx_starknet_resources = StarknetResources::default();
+    let empty_tx_gas_usage_vector =
+        empty_tx_starknet_resources.to_gas_vector(&versioned_constants, use_kzg_da);
     assert_eq!(empty_tx_gas_usage_vector, GasVector::default());
 
     // Declare.
@@ -75,13 +70,8 @@ fn test_calculate_tx_gas_usage_basic<'a>(#[values(false, true)] use_kzg_da: bool
             );
         let manual_gas_vector =
             GasVector { l1_gas: code_gas_cost.to_integer(), ..Default::default() };
-        let declare_gas_usage_vector = ActualCostBuilder::calculate_tx_gas_usage_vector(
-            &versioned_constants,
-            std::iter::empty(),
-            &declare_tx_starknet_resources,
-            use_kzg_da,
-        )
-        .unwrap();
+        let declare_gas_usage_vector =
+            declare_tx_starknet_resources.to_gas_vector(&versioned_constants, use_kzg_da);
         assert_eq!(manual_gas_vector, declare_gas_usage_vector);
     }
 
@@ -97,7 +87,7 @@ fn test_calculate_tx_gas_usage_basic<'a>(#[values(false, true)] use_kzg_da: bool
     // Manual calculation.
     let calldata_length = 0;
     let signature_length = 2;
-    let deploy_tx_starknet_resources = StarknetResources::new(
+    let deploy_account_tx_starknet_resources = StarknetResources::new(
         calldata_length,
         signature_length,
         None,
@@ -111,15 +101,10 @@ fn test_calculate_tx_gas_usage_basic<'a>(#[values(false, true)] use_kzg_da: bool
             * u128_from_usize(calldata_length + signature_length);
     let manual_starknet_gas_usage = calldata_and_signature_gas_cost.to_integer();
     let manual_gas_vector = GasVector { l1_gas: manual_starknet_gas_usage, ..Default::default() }
-        + deploy_tx_starknet_resources.get_state_changes_cost(use_kzg_da);
+        + deploy_account_tx_starknet_resources.get_state_changes_cost(use_kzg_da);
 
-    let deploy_account_gas_usage_vector = ActualCostBuilder::calculate_tx_gas_usage_vector(
-        &versioned_constants,
-        std::iter::empty(),
-        &deploy_tx_starknet_resources,
-        use_kzg_da,
-    )
-    .unwrap();
+    let deploy_account_gas_usage_vector =
+        deploy_account_tx_starknet_resources.to_gas_vector(&versioned_constants, use_kzg_da);
     assert_eq!(manual_gas_vector, deploy_account_gas_usage_vector);
 
     // L1 handler.
@@ -134,13 +119,8 @@ fn test_calculate_tx_gas_usage_basic<'a>(#[values(false, true)] use_kzg_da: bool
         std::iter::empty(),
     )
     .unwrap();
-    let l1_handler_gas_usage_vector = ActualCostBuilder::calculate_tx_gas_usage_vector(
-        &versioned_constants,
-        std::iter::empty(),
-        &l1_handler_tx_starknet_resources,
-        use_kzg_da,
-    )
-    .unwrap();
+    let l1_handler_gas_usage_vector =
+        l1_handler_tx_starknet_resources.to_gas_vector(&versioned_constants, use_kzg_da);
 
     // Manual calculation.
     let message_segment_length = get_message_segment_length(&[], Some(l1_handler_payload_size));
@@ -204,13 +184,8 @@ fn test_calculate_tx_gas_usage_basic<'a>(#[values(false, true)] use_kzg_da: bool
     )
     .unwrap();
 
-    let l2_to_l1_messages_gas_usage_vector = ActualCostBuilder::calculate_tx_gas_usage_vector(
-        &versioned_constants,
-        call_infos_iter.clone(),
-        &l2_to_l1_starknet_resources,
-        use_kzg_da,
-    )
-    .unwrap();
+    let l2_to_l1_messages_gas_usage_vector =
+        l2_to_l1_starknet_resources.to_gas_vector(&versioned_constants, use_kzg_da);
 
     // Manual calculation.
     let message_segment_length = get_message_segment_length(&l2_to_l1_payload_lengths, None);
@@ -251,13 +226,9 @@ fn test_calculate_tx_gas_usage_basic<'a>(#[values(false, true)] use_kzg_da: bool
         std::iter::empty(),
     )
     .unwrap();
-    let storage_writings_gas_usage_vector = ActualCostBuilder::calculate_tx_gas_usage_vector(
-        &versioned_constants,
-        std::iter::empty(),
-        &storage_writes_starknet_resources,
-        use_kzg_da,
-    )
-    .unwrap();
+
+    let storage_writings_gas_usage_vector =
+        storage_writes_starknet_resources.to_gas_vector(&versioned_constants, use_kzg_da);
 
     // Manual calculation.
     let manual_gas_computation =
@@ -283,13 +254,8 @@ fn test_calculate_tx_gas_usage_basic<'a>(#[values(false, true)] use_kzg_da: bool
     )
     .unwrap();
 
-    let gas_usage_vector = ActualCostBuilder::calculate_tx_gas_usage_vector(
-        &versioned_constants,
-        call_infos_iter,
-        &combined_cases_starknet_resources,
-        use_kzg_da,
-    )
-    .unwrap();
+    let gas_usage_vector =
+        combined_cases_starknet_resources.to_gas_vector(&versioned_constants, use_kzg_da);
 
     // Manual calculation.
     let fee_balance_discount = match use_kzg_da {
@@ -358,13 +324,8 @@ fn test_calculate_tx_gas_usage(#[values(false, true)] use_kzg_da: bool) {
     )
     .unwrap();
 
-    let gas_vector = ActualCostBuilder::calculate_tx_gas_usage_vector(
-        versioned_constants,
-        std::iter::empty(),
-        &starknet_resources,
-        use_kzg_da,
-    )
-    .unwrap();
+    let gas_vector = starknet_resources.to_gas_vector(versioned_constants, use_kzg_da);
+
     let GasVector { l1_gas: l1_gas_usage, l1_data_gas: l1_blob_gas_usage } = gas_vector;
     assert_eq!(u128_from_usize(tx_execution_info.actual_resources.gas_usage()), l1_gas_usage);
     assert_eq!(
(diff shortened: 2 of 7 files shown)


dorimedini-starkware (changes requested):
> Reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1646)* status: all files reviewed, 6 unresolved discussions (waiting on @nimrod-starkware)

___
*[`crates/blockifier/src/fee/actual_cost.rs` line 175 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1646#-NsNYyZx9d65S_LTXlgv:-NsNYyZx9d65S_LTXlgw:b-llv8k5) ([raw file](https://github.com/starkware-libs/blockifier/blob/977b826f0e81b78e43f8372976b98cdb92762af1/crates/blockifier/src/fee/actual_cost.rs#L175)):*
> ```XML
>         let mut actual_resources = calculate_tx_resources(
>             &self.tx_context.block_context.versioned_constants,
>             execution_resources,
> ```

what does this contain that StarknetResources doesn't?
this seems like an orthogonal structure that should be removed



_Code quote:_
```XML
execution_resources
```
___
*[`crates/blockifier/src/transaction/objects.rs` line 265 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1646#-NsN_Rxa0tGULbZeaFT_:-NsN_Rxa0tGULbZeaFTa:bj51xln) ([raw file](https://github.com/starkware-libs/blockifier/blob/977b826f0e81b78e43f8372976b98cdb92762af1/crates/blockifier/src/transaction/objects.rs#L265)):*
> ```XML
>     signature_length: usize,
>     code_size: usize,
>     events_cost: u128,
> ```

I think this is too abstract.
1. this cost is in units of gas; gas is not a starknet resource.
2. from [the computation](https://github.com/starkware-libs/blockifier/blob/aner/test_get_actual_revertible_state_changes_with_base_cache_solution/crates/blockifier/src/fee/gas_usage.rs#L31-L49), it seems like it makes sense to keep key sizes and data sizes of events.

I see two options:
1. Store two fields: `total_event_keys` and `total_event_data_size`
2. Store a vector of `StarknetEventResources` structs (each struct containing fields `key_len` and `data_len` or something)

you can then use `versioned_constants` to convert these values to gas.

I would go with option (1) to prevent allocating a large object



_Code quote:_
```XML
events_cost: u128,
```
___
*[`crates/blockifier/src/transaction/objects.rs` line 284 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1646#-NsNZstt40B74yAB5r18:-NsNZstt40B74yAB5r19:b-62zlyu) ([raw file](https://github.com/starkware-libs/blockifier/blob/977b826f0e81b78e43f8372976b98cdb92762af1/crates/blockifier/src/transaction/objects.rs#L284)):*
> ```XML
>             l1_handler_payload_size,
>             message_cost_info: MessageL1CostInfo::calculate(call_infos, l1_handler_payload_size)?,
>             events_cost: 0,
> ```

if you get `call_infos` as input, this is misleading.
from my suggestion above, if you keep key length and data length instead of costs, you can compute these values in `new` (without versioned_constants). please do the above refactor and then compute the values here



_Code quote:_
```XML
events_cost: 0,
```
___
*[`crates/blockifier/src/transaction/objects.rs` line 297 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1646#-NsNb-N03atO2A4di81B:-NsNb-N03atO2A4di81C:biyekr1) ([raw file](https://github.com/starkware-libs/blockifier/blob/977b826f0e81b78e43f8372976b98cdb92762af1/crates/blockifier/src/transaction/objects.rs#L297)):*
> ```XML
>             + self.get_code_cost(versioned_constants)
>             + self.get_state_changes_cost(use_kzg_da)
>             + self.get_messages_cost()
> ```

1. missing `get_events_cost`
2. no tests failed; this means a test is missing. please add a test that (a) fails before you add `get_events_cost` call and (b) passes when you add it
3. ... unless I am missing something here in the call stack?



_Code quote:_
```XML
        self.get_calldata_and_signature_cost(versioned_constants)
            + self.get_code_cost(versioned_constants)
            + self.get_state_changes_cost(use_kzg_da)
            + self.get_messages_cost()
```
___
*[`crates/blockifier/src/transaction/objects.rs` line 326 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1646#-NsNaHtAD80twbQCUKog:-NsNaHtAD80twbQCUKoh:b-bdyoj3) ([raw file](https://github.com/starkware-libs/blockifier/blob/977b826f0e81b78e43f8372976b98cdb92762af1/crates/blockifier/src/transaction/objects.rs#L326)):*
> ```XML
>     ) {
>         self.events_cost = call_infos
>             .map(|call_info| get_events_gas_cost(&call_info.execution.events, versioned_constants))
> ```

This function should be deleted or replaced with a more abstract computation (see above)



_Code quote:_
```XML
get_events_gas_cost
```
___
*[`crates/blockifier/src/transaction/objects.rs` line 375 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1646#-NsNaOTR7JMDfZlunFdt:-NsNaOTR7JMDfZlunFdu:bdmbti7) ([raw file](https://github.com/starkware-libs/blockifier/blob/977b826f0e81b78e43f8372976b98cdb92762af1/crates/blockifier/src/transaction/objects.rs#L375)):*
> ```XML
>     /// Returns the gas cost of the transaction's emmited events.
>     pub fn get_events_cost(&self) -> GasVector {
>         GasVector::from_l1_gas(self.events_cost)
> ```

here is where you would use versioned_constants (see above)



_Code quote:_
```XML
GasVector::from_l1_gas(self.events_cost)
```

dorimedini-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1646)* status: all files reviewed, 6 unresolved discussions (waiting on @nimrod-starkware)

___
*[`crates/blockifier/src/fee/actual_cost.rs` line 175 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1646#-NsNYyZx9d65S_LTXlgv:-NsNdRwlA78zNqr1JxUs:b-4y1hr9) ([raw file](https://github.com/starkware-libs/blockifier/blob/977b826f0e81b78e43f8372976b98cdb92762af1/crates/blockifier/src/fee/actual_cost.rs#L175)):*

what does this contain that StarknetResources doesn't?
this seems like an orthogonal structure that should be removed
</blockquote></details>

Note that this is of type `ExecutionResources` (externally... it's actually `ExecutionResourcesDef` I think), which is exactly VM resources; but `actual_resources` is of type `ResourcesMapping`.
1. just by looking at [the `impl`](https://github.com/starkware-libs/blockifier/blob/aner/test_get_actual_revertible_state_changes_with_base_cache_solution/crates/blockifier/src/transaction/objects.rs#L239-L254), we can see `ResourcesMapping` is a mix of more than one abstraction (n_steps (VM) and gas (gas) in the same dict). @Yoni-Starkware is there a reason we return `actual_resources` as a flat map like this?
2. @Yoni-Starkware if this is only for the bouncer, can we start migrating to using BouncerInfo instead of ResourcesMapping for this use case? this is a better abstraction IMO and may make ResourcesMapping redundant. We can choose to massage the output to it's original flat-map in the bridge to python if we don't want to change reponse objects.
3. Or, we could change response objects to see L2 resources (StarknetResources) in detail, along with VM resources...?
4. @nimrod-starkware PTAL at where we use ResourcesMapping other than to output the `actual_resources` in the `TransactionExecutionInfo` struct; if there are other places we need it I would like to know.

dorimedini-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1646)* status: all files reviewed, 5 unresolved discussions (waiting on @nimrod-starkware)

___
*[`crates/blockifier/src/fee/actual_cost.rs` line 175 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1646#-NsNYyZx9d65S_LTXlgv:-NsOB-TV3GeRtE_P4M74:bj0ve6l) ([raw file](https://github.com/starkware-libs/blockifier/blob/977b826f0e81b78e43f8372976b98cdb92762af1/crates/blockifier/src/fee/actual_cost.rs#L175)):*

Note that this is of type `ExecutionResources` (externally... it's actually `ExecutionResourcesDef` I think), which is exactly VM resources; but `actual_resources` is of type `ResourcesMapping`.
1. just by looking at [the `impl`](https://github.com/starkware-libs/blockifier/blob/aner/test_get_actual_revertible_state_changes_with_base_cache_solution/crates/blockifier/src/transaction/objects.rs#L239-L254), we can see `ResourcesMapping` is a mix of more than one abstraction (n_steps (VM) and gas (gas) in the same dict). @Yoni-Starkware is there a reason we return `actual_resources` as a flat map like this?
2. @Yoni-Starkware if this is only for the bouncer, can we start migrating to using BouncerInfo instead of ResourcesMapping for this use case? this is a better abstraction IMO and may make ResourcesMapping redundant. We can choose to massage the output to it's original flat-map in the bridge to python if we don't want to change reponse objects.
3. Or, we could change response objects to see L2 resources (StarknetResources) in detail, along with VM resources...?
4. @nimrod-starkware PTAL at where we use ResourcesMapping other than to output the `actual_resources` in the `TransactionExecutionInfo` struct; if there are other places we need it I would like to know.
</blockquote></details>

Answer from Yoni:
This is only for the bouncer. @nimrod-starkware we should be able to remove usage of ResourcesMapping completely, by doing option (2) above (let's talk f2f about it).

dorimedini-starkware (changes requested):
> Reviewed 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1646)* status: all files reviewed, 2 unresolved discussions (waiting on @nimrod-starkware)

___
*[`crates/blockifier/src/execution/call_info.rs` line 235 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1646#-NsOVgaY1B6_jCDfzul-:-NsOVgaY1B6_jCDfzul0:bhhc4oq) ([raw file](https://github.com/starkware-libs/blockifier/blob/f8559ae8d06c3d0022f8f568ed5ca690b8f9449e/crates/blockifier/src/execution/call_info.rs#L235)):*
> ```XML
>     }
> }
> #[derive(Clone)]
> ```

why is this needed?
`impl<'a> Iterator for CallInfoIter<'a>` should be enough, no?
this adds a new `derive` implementation that is not necessarily efficient, can't this be avoided?



_Code quote:_
```XML
#[derive(Clone)]
```

nimrod-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1646)* status: 1 of 7 files reviewed, 2 unresolved discussions (waiting on @dorimedini-starkware)

___
*[`crates/blockifier/src/execution/call_info.rs` line 235 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1646#-NsOVgaY1B6_jCDfzul-:-NsOemanCciQeL7lx6H8:b-oicyv6) ([raw file](https://github.com/starkware-libs/blockifier/blob/f8559ae8d06c3d0022f8f568ed5ca690b8f9449e/crates/blockifier/src/execution/call_info.rs#L235)):*

why is this needed?
`impl<'a> Iterator for CallInfoIter<'a>` should be enough, no?
this adds a new `derive` implementation that is not necessarily efficient, can't this be avoided?
</blockquote></details>

removed it.

dorimedini-starkware (changes requested):
> Reviewed 6 of 6 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1646)* status: all files reviewed, 3 unresolved discussions (waiting on @nimrod-starkware)

___
*[`crates/blockifier/src/execution/call_info.rs` line 235 at r4](https://reviewable.io/reviews/starkware-libs/blockifier/1646#-NsPAkFh1iMdznbciRDP:-NsPAkFh1iMdznbciRDQ:bwmy4yc) ([raw file](https://github.com/starkware-libs/blockifier/blob/b2ce06861bd13bdaadbb80f2e6b0adc3f3fe45d9/crates/blockifier/src/execution/call_info.rs#L235)):*
> ```XML
>     }
> }
> ```

restore newline (so this file isn't diffed
___
*[`crates/blockifier/src/fee/gas_usage_test.rs` line 27 at r4](https://reviewable.io/reviews/starkware-libs/blockifier/1646#-NsOoenbBOvQROCCKLOW:-NsOoenbBOvQROCCKLOX:b5ja1ow) ([raw file](https://github.com/starkware-libs/blockifier/blob/cda9e6bb3fdc356be537d52261e37da13b27243d/crates/blockifier/src/fee/gas_usage_test.rs#L27)):*
> ```XML
>     let call_info_2 = &CallInfo::default();
>     let call_info_3 = &CallInfo::default();
>     let call_infos = Some(call_info_1)
> ```

why not `.iter()`?



_Code quote:_
```XML
Some(call_info_1)
```

dorimedini-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1646)* status: all files reviewed, 3 unresolved discussions (waiting on @nimrod-starkware)

___
*[`crates/blockifier/src/transaction/objects.rs` line 297 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1646#-NsNb-N03atO2A4di81B:-NsPVFLQE_DkVAt59h4q:b-6kk3t2) ([raw file](https://github.com/starkware-libs/blockifier/blob/977b826f0e81b78e43f8372976b98cdb92762af1/crates/blockifier/src/transaction/objects.rs#L297)):*
<details><summary><i>Previously, nimrod-starkware wrote…</i></summary><blockquote>

I agree, you didn't miss anything.
Some tests tested gas vector calculation functions (for example `get_code_gas_cost`, `get_calldata_and_signature_cost`...)
At all the previous PR's, I had to edit the test and to create a `StarknetResources` instance and calculate the gas vector from that instance.
There were some cases that the test was interested in a specific resource (an [example](https://github.com/starkware-libs/blockifier/pull/1646/files#diff-c60befa67313a64cd2f9ad156230755f2c7b08840d587adc35eabb224651da72R31)), my suggestion is to make all `get_x_gas_cost` methods in the `StarknetResources` struct to private. The only way to get costs is through the `to_gas_vector` public method.
What do you think?

I will add such test as well.
</blockquote></details>

blocking until test

nimrod-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1646)* status: 1 of 8 files reviewed, 2 unresolved discussions (waiting on @dorimedini-starkware)

___
*[`crates/blockifier/src/transaction/objects.rs` line 297 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1646#-NsNb-N03atO2A4di81B:-NsbIZDt-c2qJM62BnpD:b-j33l1e) ([raw file](https://github.com/starkware-libs/blockifier/blob/977b826f0e81b78e43f8372976b98cdb92762af1/crates/blockifier/src/transaction/objects.rs#L297)):*

blocking until test
</blockquote></details>

i edited the `test_get_event_gas_cost` test to get `events_resources` from a `StarknetResources` instance. this test fails without that line that I've missed and passes with. 
___
*[`crates/blockifier/src/execution/call_info.rs` line 235 at r4](https://reviewable.io/reviews/starkware-libs/blockifier/1646#-NsPAkFh1iMdznbciRDP:-Nsb5zwg805TOn1wFsZ1:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/b2ce06861bd13bdaadbb80f2e6b0adc3f3fe45d9/crates/blockifier/src/execution/call_info.rs#L235)):*

restore newline (so this file isn't diffed
</blockquote></details>

Done.

dorimedini-starkware (changes requested):
> Reviewed 5 of 7 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1646)* status: all files reviewed, 3 unresolved discussions (waiting on @nimrod-starkware)

___
*[`crates/blockifier/src/fee/gas_usage_test.rs` line 30 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1646#-Nsbk3avCQ_CmKHrOWGE:-Nsbk3avCQ_CmKHrOWGF:b-vjo2fe) ([raw file](https://github.com/starkware-libs/blockifier/blob/d1f7b4a54ee4da7fc54fb84c7c559b7c2234471a/crates/blockifier/src/fee/gas_usage_test.rs#L30)):*
> ```XML
>     let call_info_2 = Some(CallInfo::default());
>     let call_info_3 = Some(CallInfo::default());
>     let call_infos = call_info_1.iter().chain(call_info_2.iter()).chain(call_info_3.iter());
> ```

isn't this what you're trying to do...?
why do you need the `Some`s?
you can call `call_infos.iter()` when you need an iterator



_Suggestion:_
```XML
    let call_infos = vec![CallInfo::default(), CallInfo::default(), CallInfo::default()];
```
___
*[`crates/blockifier/src/fee/gas_usage_test.rs` line 67 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1646#-NsbkM2C3JZG-RpLKeHY:-NsbkM2C3JZG-RpLKeHZ:b-atm4za) ([raw file](https://github.com/starkware-libs/blockifier/blob/d1f7b4a54ee4da7fc54fb84c7c559b7c2234471a/crates/blockifier/src/fee/gas_usage_test.rs#L67)):*
> ```XML
>         ..Default::default()
>     });
>     let call_infos = call_info_1.iter().chain(call_info_2.iter()).chain(call_info_3.iter());
> ```

why `Some`s? can't `call_infos` just be a vector of `CallInfo`s?



_Suggestion:_
```XML
    let call_info_1 = CallInfo {
        execution: CallExecution {
            events: vec![create_event(1, 2), create_event(1, 2)],
            ..Default::default()
        },
        ..Default::default()
    };
    let call_info_2 = CallInfo {
        execution: CallExecution {
            events: vec![create_event(1, 0), create_event(0, 1)],
            ..Default::default()
        },
        ..Default::default()
    };
    let call_info_3 = CallInfo {
        execution: CallExecution { events: vec![create_event(0, 1)], ..Default::default() },
        inner_calls: vec![CallInfo {
            execution: CallExecution { events: vec![create_event(1, 0)], ..Default::default() },
            ..Default::default()
        }],
        ..Default::default()
    };
    let call_infos = vec![call_info_1, call_info_2, call_info_3].iter();
```
___
*[`crates/blockifier/src/transaction/objects.rs` line 416 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1646#-NsbkpXl3UbmfnWmIblK:-NsbkpXl3UbmfnWmIblL:b-5jrfd6) ([raw file](https://github.com/starkware-libs/blockifier/blob/d1f7b4a54ee4da7fc54fb84c7c559b7c2234471a/crates/blockifier/src/transaction/objects.rs#L416)):*
> ```XML
>             })
>             .fold((0, 0), tuple_add)
>     }
> ```

if this is only used once, please meld this logic into `set_events_and_messages_resources` (it is private anyway, no need for two functions ATM).



_Code quote:_
```XML
    /// Private and static method that calculates the total_event_data_size and total_event_keys for
    /// from call_infos of a transaction.
    fn calculate_events_resources<'a>(
        call_infos: impl Iterator<Item = &'a CallInfo>,
    ) -> (u128, u128) {
        let tuple_add = |(a, b): (u128, u128), (c, d): (u128, u128)| (a + c, b + d);
        call_infos
            .map(|call_info| {
                call_info
                    .execution
                    .events
                    .iter()
                    .map(|OrderedEvent { event, .. }| {
                        // TODO(barak: 18/03/2024): Once we start charging per byte change to
                        // num_bytes_keys and num_bytes_data.
                        (u128_from_usize(event.keys.len()), u128_from_usize(event.data.0.len()))
                    })
                    .fold((0, 0), tuple_add)
            })
            .fold((0, 0), tuple_add)
    }
```

nimrod-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1646)* status: all files reviewed, 3 unresolved discussions (waiting on @dorimedini-starkware)

___
*[`crates/blockifier/src/fee/gas_usage_test.rs` line 30 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1646#-Nsbk3avCQ_CmKHrOWGE:-Nsc0rXK3gEsYZ30bnN_:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/d1f7b4a54ee4da7fc54fb84c7c559b7c2234471a/crates/blockifier/src/fee/gas_usage_test.rs#L30)):*

isn't this what you're trying to do...?
why do you need the `Some`s?
you can call `call_infos.iter()` when you need an iterator
</blockquote></details>

Done.
___
*[`crates/blockifier/src/fee/gas_usage_test.rs` line 67 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1646#-NsbkM2C3JZG-RpLKeHY:-Nsc0qxq6OKGJImCRs_L:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/d1f7b4a54ee4da7fc54fb84c7c559b7c2234471a/crates/blockifier/src/fee/gas_usage_test.rs#L67)):*

why `Some`s? can't `call_infos` just be a vector of `CallInfo`s?
</blockquote></details>

Done.
___
*[`crates/blockifier/src/transaction/objects.rs` line 416 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1646#-NsbkpXl3UbmfnWmIblK:-Nsc1ZR67tNkcQCbtah2:b8ztcia) ([raw file](https://github.com/starkware-libs/blockifier/blob/d1f7b4a54ee4da7fc54fb84c7c559b7c2234471a/crates/blockifier/src/transaction/objects.rs#L416)):*

if this is only used once, please meld this logic into `set_events_and_messages_resources` (it is private anyway, no need for two functions ATM).
</blockquote></details>

it's used twice, at the constructor and at `set_events_and_messages_resources`.
the problem is that if i move the logic to `set_events_and_messages_resources`, which takes a `&mut self`'  then i will have to return from the constructor `&mut Self` and i want to return Self from the constructor. i tried to find a solution for that problem online but couldn't find one.
ideally, the nicest way to do that is simply call `set_events_and_messages_resources(call_infos)` at the constructor.
did you encounter this?

nimrod-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1646)* status: 7 of 8 files reviewed, 3 unresolved discussions (waiting on @dorimedini-starkware)

___
*[`crates/blockifier/src/transaction/objects.rs` line 416 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1646#-NsbkpXl3UbmfnWmIblK:-NscJ_O89cCkgoTa9Yzv:bi8zd53) ([raw file](https://github.com/starkware-libs/blockifier/blob/d1f7b4a54ee4da7fc54fb84c7c559b7c2234471a/crates/blockifier/src/transaction/objects.rs#L416)):*
<details><summary><i>Previously, nimrod-starkware wrote…</i></summary><blockquote>

it's used twice, at the constructor and at `set_events_and_messages_resources`.
the problem is that if i move the logic to `set_events_and_messages_resources`, which takes a `&mut self`'  then i will have to return from the constructor `&mut Self` and i want to return Self from the constructor. i tried to find a solution for that problem online but couldn't find one.
ideally, the nicest way to do that is simply call `set_events_and_messages_resources(call_infos)` at the constructor.
did you encounter this?
</blockquote></details>

nevermind. i found a way to do that

dorimedini-starkware (approved):
:lgtm:

> Reviewed 1 of 1 files at r9, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1646#-:-NscqtdB3WZ78qkw_8GA:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nimrod-starkware)

PR #1700, opened 2024-03-21
feat(concurrency): add next_task method to the scheduler

avi-starkware:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1700)

### DIFF
--- a/crates/blockifier/src/concurrency/scheduler.rs
+++ b/crates/blockifier/src/concurrency/scheduler.rs
@@ -1,4 +1,5 @@
-use std::sync::atomic::{AtomicUsize, Ordering};
+use std::cmp::min;
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::Mutex;
 
 use crate::concurrency::TxIndex;
@@ -20,6 +21,9 @@ pub struct Scheduler {
     n_active_tasks: AtomicUsize,
     chunk_size: usize,
     tx_statuses: Box<[Mutex<TransactionStatus>]>,
+    /// Updated by the `check_done` procedure, providing a cheap way for all threads to exit their
+    /// main loops.
+    done_marker: AtomicBool,
 }
 
 // TODO(Avi, 01/04/2024): Remove dead_code attribute.
@@ -35,12 +39,13 @@ impl Scheduler {
             tx_statuses: std::iter::repeat_with(|| Mutex::new(TransactionStatus::ReadyToExecute))
                 .take(chunk_size)
                 .collect(),
+            done_marker: AtomicBool::new(false),
         }
     }
 
     /// Returns the done marker.
     pub fn done(&self) -> bool {
-        todo!()
+        self.done_marker.load(Ordering::Acquire)
     }
 
     /// Checks if all transactions have been executed and validated.
@@ -48,8 +53,29 @@ impl Scheduler {
         todo!()
     }
 
-    pub fn next_task() -> Task {
-        todo!()
+    pub fn next_task(&self) -> Task {
+        if self.done() {
+            return Task::Done;
+        }
+
+        let index_to_validate = self.validation_index.load(Ordering::Acquire);
+        let index_to_execute = self.execution_index.load(Ordering::Acquire);
+
+        if min(index_to_validate, index_to_execute) >= self.chunk_size {
+            return Task::NoTask;
+        }
+
+        if index_to_validate < index_to_execute {
+            if let Some(tx_index) = self.next_version_to_validate() {
+                return Task::ValidationTask(tx_index);
+            }
+        }
+
+        if let Some(tx_index) = self.next_version_to_execute() {
+            return Task::ExecutionTask(tx_index);
+        }
+
+        Task::NoTask
     }
 
     // TODO(barak, 01/04/2024): Ensure documentation matches logic.


barak-b-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 23 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1700#-Nu2waoR6Vk-r0HNsV1z:-Nu2waoR6Vk-r0HNsV2-:banmhea) ([raw file](https://github.com/starkware-libs/blockifier/blob/1e3f7a27a5a41c488897efb11efbff70200a1662/crates/blockifier/src/concurrency/scheduler.rs#L23)):*
> ```XML
>     pub chunk_size: u32,
>     // A marker indicating that all transactions have been executed and validated.
>     pub done_marker: AtomicBool,
> ```

IMO the name of the field doesn't need any other explanation, the interesting point that should be explained here is why we have both `check_done` and this field.



_Suggestion:_
```XML
    // Updated by the `check_done` procedure, providing a cheap way for all threads to exit their main loops.
    pub done_marker: AtomicBool,
```

barak-b-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 34 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1700#-Nu2vVhqBC_c21MjM0kg:-Nu2vVhqBC_c21MjM0kh:byzhgb3) ([raw file](https://github.com/starkware-libs/blockifier/blob/1e3f7a27a5a41c488897efb11efbff70200a1662/crates/blockifier/src/concurrency/scheduler.rs#L34)):*
> ```XML
>     pub fn done(&self) -> bool {
>         self.done_marker.load(Ordering::Acquire)
>     }
> ```

In the main loop in the thread logic, if we can access the variable without the function, I think it's cleaner.
Something like `while !scheduler.done {}` instead of `while !scheduler.done() {}`.
This way, we avoid this "redundant" code which is very similar to property in Python, just that I can't see why to use it.
Maybe the atomicity mechanism will not allow doing what I suggested (and then it's a good reason to do that).



_Code quote:_
```XML
    // Returns the done marker.
    pub fn done(&self) -> bool {
        self.done_marker.load(Ordering::Acquire)
    }
```

avi-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1700)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @barak-b-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 23 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1700#-Nu2waoR6Vk-r0HNsV1z:-Nu4LrPK18vohVyJUbcT:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/1e3f7a27a5a41c488897efb11efbff70200a1662/crates/blockifier/src/concurrency/scheduler.rs#L23)):*

IMO the name of the field doesn't need any other explanation, the interesting point that should be explained here is why we have both `check_done` and this field.
</blockquote></details>

Done.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 34 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1700#-Nu2vVhqBC_c21MjM0kg:-Nu4MPZZ0GApQRSuT6cB:bcguu1n) ([raw file](https://github.com/starkware-libs/blockifier/blob/1e3f7a27a5a41c488897efb11efbff70200a1662/crates/blockifier/src/concurrency/scheduler.rs#L34)):*

In the main loop in the thread logic, if we can access the variable without the function, I think it's cleaner.
Something like `while !scheduler.done {}` instead of `while !scheduler.done() {}`.
This way, we avoid this "redundant" code which is very similar to property in Python, just that I can't see why to use it.
Maybe the atomicity mechanism will not allow doing what I suggested (and then it's a good reason to do that).
</blockquote></details>

I don't think it is good practice to touch the atomic variables of the scheduler from outside the scheduler.
The orderings are confusing enough when they are all in the same file.
Plus, it is an implementation detail that shouldn't matter to the worker running the code.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 44 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1700#-NtzKc_k5vzqfSqt0OeT:-Nu4N9TO34QANPXVuWFF:b540s6z) ([raw file](https://github.com/starkware-libs/blockifier/blob/1e3f7a27a5a41c488897efb11efbff70200a1662/crates/blockifier/src/concurrency/scheduler.rs#L44)):*

Why is the `loop` needed?
Is it better to have it here than in the thread itself?
</blockquote></details>

This is not the loop of the task itself. This loop is just to continue trying to find a task until something is returned. In most cases, it shouldn't run more than one iteration.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 62 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1700#-NtzLei5DdUZPrVIQ_BB:-Nu4PJFR9p550Y9SA-Ar:biymsf2) ([raw file](https://github.com/starkware-libs/blockifier/blob/1e3f7a27a5a41c488897efb11efbff70200a1662/crates/blockifier/src/concurrency/scheduler.rs#L62)):*

Do we want to run an execution task in case of`idx_to_validate < idx_to_execute` but no relevant transaction is ready to be validated?
</blockquote></details>

You are right. Let's stick to the algorithm in the paper in these small details.
(I based this deviation from the algorithm on the aptos repo.)

barak-b-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 34 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1700#-Nu2vVhqBC_c21MjM0kg:-Nu4X0WDExw3WJodzdqi:bh9n7po) ([raw file](https://github.com/starkware-libs/blockifier/blob/1e3f7a27a5a41c488897efb11efbff70200a1662/crates/blockifier/src/concurrency/scheduler.rs#L34)):*

I don't think it is good practice to touch the atomic variables of the scheduler from outside the scheduler.
The orderings are confusing enough when they are all in the same file.
Plus, it is an implementation detail that shouldn't matter to the worker running the code.
</blockquote></details>

You're right (: resolved.

noaov1 (changes requested):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1700)* status: all files reviewed, 4 unresolved discussions (waiting on @avi-starkware and @barak-b-starkware)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 44 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1700#-NtzKc_k5vzqfSqt0OeT:-NvQSbuy6Hco4Z1NK3KX:b-zh1m4j) ([raw file](https://github.com/starkware-libs/blockifier/blob/1e3f7a27a5a41c488897efb11efbff70200a1662/crates/blockifier/src/concurrency/scheduler.rs#L44)):*

This is not the loop of the task itself. This loop is just to continue trying to find a task until something is returned. In most cases, it shouldn't run more than one iteration.
</blockquote></details>

Why do we need it in addition to the thread loop?
When will we have more than one iteration?
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 62 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1700#-NtzLei5DdUZPrVIQ_BB:-NvQUfI16gPTJrI--ImY:b-hmcmzk) ([raw file](https://github.com/starkware-libs/blockifier/blob/1e3f7a27a5a41c488897efb11efbff70200a1662/crates/blockifier/src/concurrency/scheduler.rs#L62)):*

You are right. Let's stick to the algorithm in the paper in these small details.
(I based this deviation from the algorithm on the aptos repo.)
</blockquote></details>

Interesting. Let's think. 
Assume `idx_to_validate < idx_to_execute` and no task is ready to be validated. When it can happen?
We finish validation of `tx_i` and `tx_i+1` have not finish execution yet. Do we want to execute `tx_j` for `j` necessarily larger from `i+1`? Or wait for `tx_i+1` to be ready?
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 55 at r4](https://reviewable.io/reviews/starkware-libs/blockifier/1700#-NvQTp0Z62jPblDJHXOC:-NvQTp0Z62jPblDJHXOD:b-gzt7sl) ([raw file](https://github.com/starkware-libs/blockifier/blob/3da4dede04f26a785a637e2ddc6c9234e18e3919/crates/blockifier/src/concurrency/scheduler.rs#L55)):*
> ```XML
>             if min(index_to_validate, index_to_execute) >= self.chunk_size {
>                 return Task::NoTask;
>             }
> ```

We catch it in the `next_version_to_validate/execute` methods, no?
Why is it needed here as well? For an early return?



_Code quote:_
```XML
            if min(index_to_validate, index_to_execute) >= self.chunk_size {
                return Task::NoTask;
            }
```

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 44 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1700#-NtzKc_k5vzqfSqt0OeT:-NvQcF6XBpu9T7CwihtA:bgxoy0s) ([raw file](https://github.com/starkware-libs/blockifier/blob/1e3f7a27a5a41c488897efb11efbff70200a1662/crates/blockifier/src/concurrency/scheduler.rs#L44)):*

Why do we need it in addition to the thread loop?
When will we have more than one iteration?
</blockquote></details>

It is mainly for the case where `next_version_to_validate` / `next_version_to_execute` return `None`.
This might be due to many threads increasing the counters simultaneously, such that the counter exceeds `chunk_size` by the time the current thread finishes `next_version_to_validate` / `next_version_to_execute`.
The more common reason I think would be because the `tx_status` is not right yet.

This loop ensures that the method `next_task` always returns a task. `NoTask` is returned only when all relevant tasks to be done are executing in other threads, and `Done` is returned only when the entire chunk is ready to be committed.

It is an optimization from the aptos repo. I now see that in the case where `NoTask` is returned, the thread loop simply calls `next_task` again. I guess they just wanted to skip some of the overhead of restarting the thread loop.

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 55 at r4](https://reviewable.io/reviews/starkware-libs/blockifier/1700#-NvQTp0Z62jPblDJHXOC:-NvQbpXQ4Opdl7cgNAOi:bcfbn5) ([raw file](https://github.com/starkware-libs/blockifier/blob/3da4dede04f26a785a637e2ddc6c9234e18e3919/crates/blockifier/src/concurrency/scheduler.rs#L55)):*

We catch it in the `next_version_to_validate/execute` methods, no?
Why is it needed here as well? For an early return?
</blockquote></details>

We load these two atomic variables anyway to compare their values, so this check has minimal overhead and saves us going into `next_version_to_...` when it is clearly not needed.

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 62 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1700#-NtzLei5DdUZPrVIQ_BB:-NvQyzEA57Lk2pcilpa3:b-v6gfvu) ([raw file](https://github.com/starkware-libs/blockifier/blob/1e3f7a27a5a41c488897efb11efbff70200a1662/crates/blockifier/src/concurrency/scheduler.rs#L62)):*

Interesting. Let's think. 
Assume `idx_to_validate < idx_to_execute` and no task is ready to be validated. When it can happen?
We finish validation of `tx_i` and `tx_i+1` have not finish execution yet. Do we want to execute `tx_j` for `j` necessarily larger from `i+1`? Or wait for `tx_i+1` to be ready?
</blockquote></details>

I think we shouldn't wait, actually. Think of the first few thread loops. We'd want all the threads to start executing immediately and not have some threads stuck waiting for a validation task to be available when transactions are still being executed. (Barak's suggestion to have validation_index start at chunk_size slightly mitigates this case, but once the first execution completes, the validation_index will be decreased, and we will have threads waiting for validation tasks to be available anyway.)

noaov1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1700)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @avi-starkware and @barak-b-starkware)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 62 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1700#-NtzLei5DdUZPrVIQ_BB:-NvVOoIL3HLCmcQ4sv9-:b-fhmhzu) ([raw file](https://github.com/starkware-libs/blockifier/blob/1e3f7a27a5a41c488897efb11efbff70200a1662/crates/blockifier/src/concurrency/scheduler.rs#L62)):*

I think we shouldn't wait, actually. Think of the first few thread loops. We'd want all the threads to start executing immediately and not have some threads stuck waiting for a validation task to be available when transactions are still being executed. (Barak's suggestion to have validation_index start at chunk_size slightly mitigates this case, but once the first execution completes, the validation_index will be decreased, and we will have threads waiting for validation tasks to be available anyway.)
</blockquote></details>

Makes sense. Why then the paper choose otherwise?

noaov1 (approved):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1700)* status: all files reviewed, 2 unresolved discussions (waiting on @avi-starkware and @barak-b-starkware)

avi-starkware (commented):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1700)* status: all files reviewed, 1 unresolved discussion (waiting on @barak-b-starkware)

avi-starkware (commented):
> Dismissed @barak-b-starkware from a discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1700)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @barak-b-starkware)

PR #1702, opened 2024-03-21
feat(concurrency): add counters to the scheduler

avi-starkware:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1702)

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -21,6 +21,7 @@ const Configuration = {
      */
     rules: {
         'scope-enum': [2, 'always', [
+            'concurrency',
             'execution',
             'fee',
             'native_blockifier',
--- a/crates/blockifier/src/concurrency/scheduler.rs
+++ b/crates/blockifier/src/concurrency/scheduler.rs
@@ -1,35 +1,107 @@
+use std::sync::atomic::{AtomicU32, Ordering};
+
 use crate::concurrency::Version;
 
 #[cfg(test)]
 #[path = "scheduler_test.rs"]
 pub mod test;
 
-pub struct Scheduler;
+// TODO(Avi, 01/04/2024): Remove dead_code attribute.
+#[allow(dead_code)]
+#[derive(Default)]
+pub struct Scheduler {
+    execution_index: AtomicU32,
+    validation_index: AtomicU32,
+    /// Read twice upon checking the chunk completion. Used to detect if validation or execution
+    /// index decreased from their observed values after ensuring that the number of active tasks
+    /// is zero.
+    decrease_counter: AtomicU32,
+    n_active_tasks: AtomicU32,
+    chunk_size: u32,
+}
 
+// TODO(Avi, 01/04/2024): Remove dead_code attribute.
+#[allow(dead_code)]
 impl Scheduler {
-    pub fn done() -> bool {
+    pub fn new(chunk_size: u32) -> Scheduler {
+        Scheduler { chunk_size, validation_index: chunk_size.into(), ..Default::default() }
+    }
+
+    /// Returns the done marker.
+    pub fn done(&self) -> bool {
         todo!()
     }
 
-    /// Return the next task to run. Prioratize validations tasks.
+    /// Checks if all transactions have been executed and validated.
+    fn check_done(&self) {
+        todo!()
+    }
+
     pub fn next_task() -> Task {
         todo!()
     }
 
     // TODO(barak, 01/04/2024): Ensure documentation matches logic.
-    /// Update the Scheduler that an execution task has been finished and trigger the creation of
-    /// new tasks accordingly: schedule validation for the current and higher transactions, if
-    /// not already scheduled.
-    pub fn finish_execution() -> Task {
+    /// Updates the Scheduler that an execution task has been finished and triggers the creation of
+    /// new tasks accordingly: schedules validation for the current and higher transactions, if not
+    /// already scheduled.
+    pub fn finish_execution(&self) -> Task {
         todo!()
     }
 
     // TODO(barak, 01/04/2024): Ensure documentation matches logic.
-    /// Update the Scheduler that a validation task has been finished and trigger the creation of
-    /// new tasks accordingly: schedule validation for higher transactions + re-executes the
+    /// Updates the Scheduler that a validation task has been finished and triggers the creation of
+    /// new tasks in case of failure: schedules validation for higher transactions + re-executes the
     /// current transaction (if ready).
-    pub fn finish_validation() -> Task {
+    pub fn finish_validation(&self) -> Task {
         todo!()
+    }
+
+    fn decrease_validation_index(&self, target_idx: u32) {
+        self.validation_index.fetch_min(target_idx, Ordering::SeqCst);
+        self.decrease_counter.fetch_add(1, Ordering::SeqCst);
+    }
+
+    fn decrease_execution_index(&self, target_idx: u32) {
+        self.execution_index.fetch_min(target_idx, Ordering::SeqCst);
+        self.decrease_counter.fetch_add(1, Ordering::SeqCst);
+    }
+
+    /// Updates a transaction's status to `Executing` if it is ready to execute.
+    fn try_incarnate(&self, tx_idx: u32) -> Option<u32> {
+        if tx_idx < self.chunk_size {
+            // TODO(barak, 01/04/2024): complete try_incarnate logic.
+            return Some(tx_idx);
+        }
+        self.n_active_tasks.fetch_sub(1, Ordering::SeqCst);
+        None
+    }
+
+    fn next_version_to_validate(&self) -> Option<u32> {
+        let index_to_validate = self.validation_index.load(Ordering::Acquire);
+        if index_to_validate >= self.chunk_size {
+            self.check_done();
+            return None;
+        }
+        self.n_active_tasks.fetch_add(1, Ordering::SeqCst);
+        let index_to_validate = self.validation_index.fetch_add(1, Ordering::SeqCst);
+        if index_to_validate < self.chunk_size {
+            // TODO(barak, 01/04/2024): complete next_version_to_validate logic.
+            return Some(index_to_validate);
+        }
+        self.n_active_tasks.fetch_sub(1, Ordering::SeqCst);
+        None
+    }
+
+    fn next_version_to_execute(&self) -> Option<u32> {
+        let index_to_execute = self.execution_index.load(Ordering::Acquire);
+        if index_to_execute >= self.chunk_size {
+            self.check_done();
+            return None;
+        }
+        self.n_active_tasks.fetch_add(1, Ordering::SeqCst);
+        let index_to_execute = self.execution_index.fetch_add(1, Ordering::SeqCst);
+        self.try_incarnate(index_to_execute)
     }
 }
 


avi-starkware:
This PR is a new version of #1699 that has been closed.

noaov1 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1702)* status: all files reviewed, 4 unresolved discussions (waiting on @avi-starkware and @barak-b-starkware)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 15 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-NtjkYOz2WhQl95w0_U2:-NtjkYP-0A5vA3R3AvXL:b-hv0l16) ([raw file](https://github.com/starkware-libs/blockifier/blob/d61e33de0f40fff797c337db97e50d6aa7fb1b57/crates/blockifier/src/concurrency/scheduler.rs#L15)):*
> ```XML
>     pub n_active_tasks: AtomicU32,
>     pub n_transactions: u32,
> }
> ```

Please add a short doc string for each counter.



_Code quote:_
```XML
pub struct Scheduler {
    pub execution_idx: AtomicU32,
    pub validation_idx: AtomicU32,
    pub decrease_counter: AtomicU32,
    pub n_active_tasks: AtomicU32,
    pub n_transactions: u32,
}
```
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 79 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-NtkaLJX4Qefp7GTvoDF:-NtkaLJX4Qefp7GTvoDG:b-blaf21) ([raw file](https://github.com/starkware-libs/blockifier/blob/d61e33de0f40fff797c337db97e50d6aa7fb1b57/crates/blockifier/src/concurrency/scheduler.rs#L79)):*
> ```XML
> 
>     pub fn next_version_to_validate(&self) -> Option<u32> {
>         let idx_to_validate = self.validation_idx.load(Ordering::Acquire);
> ```

Can you please explain why you chose `Ordering::Acquire`?



_Code quote:_
```XML
let idx_to_validate = self.validation_idx.load(Ordering::Acquire);
```
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 80 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Ntkb-8O8GYRGsKcMt0g:-Ntkb-8O8GYRGsKcMt0h:b871pn) ([raw file](https://github.com/starkware-libs/blockifier/blob/d61e33de0f40fff797c337db97e50d6aa7fb1b57/crates/blockifier/src/concurrency/scheduler.rs#L80)):*
> ```XML
>     pub fn next_version_to_validate(&self) -> Option<u32> {
>         let idx_to_validate = self.validation_idx.load(Ordering::Acquire);
>         if idx_to_validate >= self.n_transactions {
> ```

Consider defining a block size constant



_Suggestion:_
```XML
if idx_to_validate >= BLOCK_SIZE
```
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 85 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-NtkcGxMEnXwklsiPTee:-NtkcGxMEnXwklsiPTef:bfg5ks2) ([raw file](https://github.com/starkware-libs/blockifier/blob/d61e33de0f40fff797c337db97e50d6aa7fb1b57/crates/blockifier/src/concurrency/scheduler.rs#L85)):*
> ```XML
>         }
>         self.n_active_tasks.fetch_add(1, Ordering::SeqCst);
>         let idx_to_validate = self.validation_idx.fetch_add(1, Ordering::SeqCst);
> ```

Can you please explain why you chose `Ordering::SeqCst`?



_Code quote:_
```XML
        self.n_active_tasks.fetch_add(1, Ordering::SeqCst);
        let idx_to_validate = self.validation_idx.fetch_add(1, Ordering::SeqCst);
```

avi-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1702)* status: all files reviewed, 3 unresolved discussions (waiting on @barak-b-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 79 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-NtkaLJX4Qefp7GTvoDF:-Ntl5SYD9Q18AOrTWKpA:b-prg9am) ([raw file](https://github.com/starkware-libs/blockifier/blob/d61e33de0f40fff797c337db97e50d6aa7fb1b57/crates/blockifier/src/concurrency/scheduler.rs#L79)):*

Can you please explain why you chose `Ordering::Acquire`?
</blockquote></details>

It's the typical ordering to be used with a load operation. As far as I understand, it makes it so that all write operations that happened before the last time `validation_idx` has been stored (in the thread that stored the value) will be ordered to happen before this store operation (i.e., all these writes that happened in the other thread, also already happened from the viewpoint of the current thread).
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 85 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-NtkcGxMEnXwklsiPTee:-Ntl6Zfv7u-CeejKDgdN:bdx7y7t) ([raw file](https://github.com/starkware-libs/blockifier/blob/d61e33de0f40fff797c337db97e50d6aa7fb1b57/crates/blockifier/src/concurrency/scheduler.rs#L85)):*

Can you please explain why you chose `Ordering::SeqCst`?
</blockquote></details>

This is the strongest ordering you can choose from. The order of all operations with this ordering will appear the same from the context of all existing threads.
I am not entirely sure why this is necessary, but the in the aptos repo they use this ordering for all operations that update one of the counters.
This is the "safest" ordering as it allows no unexpected inconsistencies between threads, but it puts the most limits on compiler optimizations - so it might hurt performance.

avi-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1702)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @barak-b-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 15 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-NtjkYOz2WhQl95w0_U2:-NtlAwwS7axCx0eX6mIk:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/d61e33de0f40fff797c337db97e50d6aa7fb1b57/crates/blockifier/src/concurrency/scheduler.rs#L15)):*

Please add a short doc string for each counter.
</blockquote></details>

Done.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 80 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Ntkb-8O8GYRGsKcMt0g:-NtlAyGZ0xhFTOFt-8AK:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/d61e33de0f40fff797c337db97e50d6aa7fb1b57/crates/blockifier/src/concurrency/scheduler.rs#L80)):*

Consider defining a block size constant
</blockquote></details>

Done.

noaov1 (changes requested):
Please add tests (in a separate PR).

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-:-NtoIybl2jzb36OFUJqc:b-vmicd9)* status: all files reviewed, 6 unresolved discussions (waiting on @avi-starkware and @barak-b-starkware)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 52 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Ntl91xuEzHuT4HN1pco:-Ntl91xv6MDcLwUvzzsn:undefined) ([raw file](https://github.com/starkware-libs/blockifier/blob/d61e33de0f40fff797c337db97e50d6aa7fb1b57/crates/blockifier/src/concurrency/scheduler.rs#L52)):*
> ```XML
>     /// Update the Scheduler that a validation task has been finished and trigger the creation of
>     /// new tasks accordingly: schedule validation for higher transactions + re-executes the
>     /// current transaction (if ready).
> ```





_Suggestion:_
```XML
    /// new tasks in case of failure: schedule validation for higher transactions + re-executes the
    /// current transaction (if ready).
```
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 76 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Ntl2t9I0dSDae8bAp1R:-Ntl2t9I0dSDae8bAp1S:ba5gwkx) ([raw file](https://github.com/starkware-libs/blockifier/blob/d61e33de0f40fff797c337db97e50d6aa7fb1b57/crates/blockifier/src/concurrency/scheduler.rs#L76)):*
> ```XML
>         self.n_active_tasks.fetch_sub(1, Ordering::SeqCst);
>         None
>     }
> ```

Can we avoid this method for now?
Q- If we can't execute the relevant transaction (the one that the execution index is pointing to), then we are not executing the next ones as well, right?
Same for validation. Why is that?



_Code quote:_
```XML
    pub fn try_incarnate(&self, tx_idx: u32) -> Option<u32> {
        if tx_idx < self.n_transactions {
            // TODO: implement tx_status
            // if tx_status == TxStatus::ReadyToExecute
            //     tx_status = TxStatus::Executing
            return Some(tx_idx);
        }
        self.n_active_tasks.fetch_sub(1, Ordering::SeqCst);
        None
    }
```
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 26 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-NtlfoPB0h8onBDIEt4Q:-NtlfoPB0h8onBDIEt4R:byzk67) ([raw file](https://github.com/starkware-libs/blockifier/blob/0e04e307625b855c96d67280ad2a8a1fdfb141d3/crates/blockifier/src/concurrency/scheduler.rs#L26)):*
> ```XML
>         Self::new()
>     }
> }
> ```

Consider add `#[derive(Default)]` above the `Scheduler` struct. 



_Code quote:_
```XML
impl Default for Scheduler {
    fn default() -> Self {
        Self::new()
    }
}
```
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 36 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-NtlfznuDdUK0Fu5YTYa:-Ntlfznv6DBBOKkRESOh:bxtub52) ([raw file](https://github.com/starkware-libs/blockifier/blob/0e04e307625b855c96d67280ad2a8a1fdfb141d3/crates/blockifier/src/concurrency/scheduler.rs#L36)):*
> ```XML
>             n_active_tasks: AtomicU32::new(0),
>         }
>     }
> ```

Is it needed?



_Code quote:_
```XML
    pub fn new() -> Scheduler {
        Scheduler {
            execution_idx: AtomicU32::new(0),
            validation_idx: AtomicU32::new(0),
            decrease_counter: AtomicU32::new(0),
            n_active_tasks: AtomicU32::new(0),
        }
    }
```

avi-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1702)* status: 0 of 1 files reviewed, 6 unresolved discussions (waiting on @barak-b-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 52 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Ntl91xuEzHuT4HN1pco:-NtpU2szDc_BUAyLOBL3:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/d61e33de0f40fff797c337db97e50d6aa7fb1b57/crates/blockifier/src/concurrency/scheduler.rs#L52)):*
> ```XML
>     /// Update the Scheduler that a validation task has been finished and trigger the creation of
>     /// new tasks accordingly: schedule validation for higher transactions + re-executes the
>     /// current transaction (if ready).
> ```

Done.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 76 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Ntl2t9I0dSDae8bAp1R:-NtpWO1y-5v3Rj8hVJlq:b-t5vg31) ([raw file](https://github.com/starkware-libs/blockifier/blob/d61e33de0f40fff797c337db97e50d6aa7fb1b57/crates/blockifier/src/concurrency/scheduler.rs#L76)):*

Can we avoid this method for now?
Q- If we can't execute the relevant transaction (the one that the execution index is pointing to), then we are not executing the next ones as well, right?
Same for validation. Why is that?
</blockquote></details>

We need this method since it will be used in `finish_validation`. I changed the name of the method to `try_schedule_execute`.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 26 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-NtlfoPB0h8onBDIEt4Q:-NtpTO1d4e3WmcHVzlro:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/0e04e307625b855c96d67280ad2a8a1fdfb141d3/crates/blockifier/src/concurrency/scheduler.rs#L26)):*

Consider add `#[derive(Default)]` above the `Scheduler` struct.
</blockquote></details>

Done.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 36 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-NtlfznuDdUK0Fu5YTYa:-NtpTQ0j8JWIuay90BH8:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/0e04e307625b855c96d67280ad2a8a1fdfb141d3/crates/blockifier/src/concurrency/scheduler.rs#L36)):*

Is it needed?
</blockquote></details>

Done.

barak-b-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 20 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2itrc0B8CQzO_YXak:-Nu2itrc0B8CQzO_YXal:b4fk35w) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L20)):*
> ```XML
>     pub n_active_tasks: AtomicU32,
>     // The number of transactions in each chunk.
>     pub chunk_size: u32,
> ```

IMO, only the `decrease_counter` necessitates an explanation, all others are self-explanatory.
Moreover, to be (kind of) consistent, WDYT about `cnt` instead of `counter` (like `idx` instead of `index`).
(And I think you wanted here `///` and not `//`?).



_Suggestion:_
```XML
    pub execution_idx: AtomicU32,
    pub validation_idx: AtomicU32,
    /// Read twice upon checking the chunk completion. Used to detect if validation or execution index decreases from their observed values after ensuring that the number of active tasks is zero.
    pub decrease_counter: AtomicU32,
    pub n_active_tasks: AtomicU32,
    pub chunk_size: u32,
```

barak-b-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 26 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2l1oo9iiqPSuXIu9P:-Nu2l1oo9iiqPSuXIu9Q:b-s99ux4) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L26)):*
> ```XML
>     pub fn new(chunk_size: u32) -> Scheduler {
>         Scheduler { chunk_size, ..Default::default() }
>     }
> ```

Not sure 100% but I think it should be like that:



_Suggestion:_
```XML
    pub fn new(chunk_size: u32) -> Scheduler {
        Scheduler { chunk_size, validation_idx: chunk_size + 1, ..Default::default() }
    }
```

barak-b-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 34 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2m3qO9wpGaDrramAn:-Nu2m3qO9wpGaDrramAo:billzr8) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L34)):*
> ```XML
> 
>     // Checks if all transactions have been executed and validated. Namely, if both the execution
>     // and validation indexes are at least the chunk size, and there are no active tasks.
> ```

IMO it's more complicated, so better to leave the documentation simpler.



_Suggestion:_
```XML
    // Checks if all transactions have been executed and validated.
```

barak-b-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 72 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2nuPA6apUlwvCkCcZ:-Nu2nuPA6apUlwvCkCc_:b-1tjs64) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L72)):*
> ```XML
>         self.execution_idx.fetch_min(target_idx, Ordering::SeqCst);
>         self.decrease_counter.fetch_add(1, Ordering::SeqCst);
>     }
> ```

IMO the function name and its two lines are self-explanatory so no need to elaborate, anyway, if you do want to leave it here, I think you want it with `///` and not `//`



_Suggestion:_
```XML
    pub fn decrease_execution_idx(&self, target_idx: u32) {
        self.execution_idx.fetch_min(target_idx, Ordering::SeqCst);
        self.decrease_counter.fetch_add(1, Ordering::SeqCst);
    }
```

barak-b-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 65 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2oOA204OQ0ujLA574:-Nu2oOA204OQ0ujLA575:b-1tjs64) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L65)):*
> ```XML
>         self.validation_idx.fetch_min(target_idx, Ordering::SeqCst);
>         self.decrease_counter.fetch_add(1, Ordering::SeqCst);
>     }
> ```

IMO the function name and its two lines are self-explanatory so no need to elaborate, anyway, if you do want to leave it here, I think you want it with `///` and not `//`



_Suggestion:_
```XML
    pub fn decrease_validation_idx(&self, target_idx: u32) {
        self.validation_idx.fetch_min(target_idx, Ordering::SeqCst);
        self.decrease_counter.fetch_add(1, Ordering::SeqCst);
    }
```

barak-b-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 85 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2qGpj1wNlzquSEtV9:-Nu2qGpj1wNlzquSEtVA:b-5xjwlq) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L85)):*
> ```XML
>         self.n_active_tasks.fetch_sub(1, Ordering::SeqCst);
>         None
>     }
> ```

I see why we don't want the "incarnate" terminology but suggest re-consider the name for two reasons:
1. Being consistent with the paper might be helpful in the future for us and for others.
2. `try_incarnate` doesn't increment the incarnation number, it just reads and returns, so as long as we don't have incarnation numbers it will just return the `tx_idx` and the translation can fit.



_Suggestion:_
```XML
    // Checks whether or not the transaction is ready to execute and updates the number of active tasks accordingly.
    pub fn try_incarnate(&self, tx_idx: u32) -> Option<u32> {
        if tx_idx < self.chunk_size {
            // TODO(barak, 01/04/2024): complete `try_incarnate` logic.
            return Some(tx_idx);
        }
        self.n_active_tasks.fetch_sub(1, Ordering::SeqCst);
        None
    }
```

barak-b-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 103 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2rXFV8VM5AZ_BTaHl:-Nu2rXFV8VM5AZ_BTaHm:bkddyc1) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L103)):*
> ```XML
>         self.n_active_tasks.fetch_sub(1, Ordering::SeqCst);
>         None
>     }
> ```

Note that `fetch_add` returns the previous value and IIUC we want here the new value (after adding one).



_Suggestion:_
```XML
    pub fn next_version_to_validate(&self) -> Option<u32> {
        let idx_to_validate = self.validation_idx.load(Ordering::Acquire);
        if idx_to_validate >= self.chunk_size {
            self.check_done();
            return None;
        }
        self.n_active_tasks.fetch_add(1, Ordering::SeqCst);
        self.validation_idx.fetch_add(1, Ordering::SeqCst);
        if self.validation_idx < self.chunk_size {
            // TODO(barak, 01/04/2024): complete `next_version_to_Validate` logic.
            return Some(idx_to_validate);
        }
        self.n_active_tasks.fetch_sub(1, Ordering::SeqCst);
        None
    }
```

barak-b-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 115 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2skg9C0uWfn0aobYU:-Nu2skg9C0uWfn0aobYV:b-68ars7) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L115)):*
> ```XML
>         let idx_to_execute = self.execution_idx.fetch_add(1, Ordering::SeqCst);
>         self.try_schedule_execute(idx_to_execute)
>     }
> ```

Same as above (previous value).



_Suggestion:_
```XML
    pub fn next_version_to_execute(&self) -> Option<u32> {
        let idx_to_execute = self.execution_idx.load(Ordering::Acquire);
        if idx_to_execute >= self.chunk_size {
            self.check_done();
            return None;
        }
        self.n_active_tasks.fetch_add(1, Ordering::SeqCst);
        self.execution_idx.fetch_add(1, Ordering::SeqCst);
        self.try_schedule_execute(self.execution_idx)
    }
```

barak-b-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 26 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2l1oo9iiqPSuXIu9P:-Nu2u1SjA-Sz8imu5wbV:btglyn1) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L26)):*

Not sure 100% but I think it should be like that:
</blockquote></details>

Or maybe add an `impl Default` for the scheduler that encapsulates the logic?

barak-b-starkware (commented):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1702)* status: all files reviewed, 11 unresolved discussions (waiting on @avi-starkware and @noaov1)

avi-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1702)* status: 0 of 2 files reviewed, 11 unresolved discussions (waiting on @barak-b-starkware and @noaov1)

___
*[`-- commits` line 2 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2uY82-68GYX4WQeLc):*
Done.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 20 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2itrc0B8CQzO_YXak:-Nu3TZoy9xm4ECQCHe-S:b-cu03hr) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L20)):*

IMO, only the `decrease_counter` necessitates an explanation, all others are self-explanatory.
Moreover, to be (kind of) consistent, WDYT about `cnt` instead of `counter` (like `idx` instead of `index`).
(And I think you wanted here `///` and not `//`?).
</blockquote></details>

Actually, to be consistent, I should change `idx` to `index`. We don't use abbreviations in our variable names (except for tx).

Fixed all comments.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 26 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2l1oo9iiqPSuXIu9P:-Nu3SYAfAwEFqVNcg30U:bivglqr) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L26)):*

Or maybe add an `impl Default` for the scheduler that encapsulates the logic?
</blockquote></details>

Why should the validation index be chunk_size + 1 when the scheduler is initialized? I think it should start at 0 
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 34 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2m3qO9wpGaDrramAn:-Nu3UejmAovr1-WEsNF2:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L34)):*

IMO it's more complicated, so better to leave the documentation simpler.
</blockquote></details>

Done.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 65 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2oOA204OQ0ujLA574:-Nu3UjB49Dr-tH5H0CwY:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L65)):*

IMO the function name and its two lines are self-explanatory so no need to elaborate, anyway, if you do want to leave it here, I think you want it with `///` and not `//`
</blockquote></details>

Done.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 72 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2nuPA6apUlwvCkCcZ:-Nu3Ujls1I81XD6_5iYJ:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L72)):*

IMO the function name and its two lines are self-explanatory so no need to elaborate, anyway, if you do want to leave it here, I think you want it with `///` and not `//`
</blockquote></details>

Done.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 85 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2qGpj1wNlzquSEtV9:-Nu3V1BJ3LOTMi8FxHlk:b-2uwi2f) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L85)):*

I see why we don't want the "incarnate" terminology but suggest re-consider the name for two reasons:
1. Being consistent with the paper might be helpful in the future for us and for others.
2. `try_incarnate` doesn't increment the incarnation number, it just reads and returns, so as long as we don't have incarnation numbers it will just return the `tx_idx` and the translation can fit.
</blockquote></details>

I don't mind either way regarding the name of the function. @noaov1 What do you think?

I made the wording of the doc slightly different than you suggested.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 103 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2rXFV8VM5AZ_BTaHl:-Nu3Wmcz-qxzV3IDdZux:bv0ksid) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L103)):*

Note that `fetch_add` returns the previous value and IIUC we want here the new value (after adding one).
</blockquote></details>

1. `self.validation_idx` does not work for reading the atomic variable's value. You have to `load` it via an atomic method.
2. We want the current `validation_idx`, not the next one. It is not very explicit in the paper, but it makes sense (e.g., if we always take the next one, we never validate/execute tx 0), and this is what they do in the aptos repo.

You are right some of the logic here is unnecessary, though, and I fixed the comments as you suggested.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 115 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2skg9C0uWfn0aobYU:-Nu3_k3C6FQt5CXtDTIi:b-som1v5) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L115)):*

Same as above (previous value).
</blockquote></details>

See the answer to your previous comment.

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 26 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2l1oo9iiqPSuXIu9P:-Nu3aarX6B0W6dki4VYI:bd5rtjy) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L26)):*

Why should the validation index be chunk_size + 1 when the scheduler is initialized? I think it should start at 0
</blockquote></details>

This is also consistent with the paper and the aptos repo.

barak-b-starkware (changes requested):
> Reviewed 1 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1702)* status: all files reviewed, 7 unresolved discussions (waiting on @avi-starkware and @noaov1)

___
*[`-- commits` line 2 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2uY82-68GYX4WQeLc):*

Done.
</blockquote></details>

Do we want to add new scope `concurrency` in addition to `execution` in the commitlint configuration? so it would be something like:
`feat(concurrency, execution): add counters to the scheduler`
@noaov1 
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 26 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2l1oo9iiqPSuXIu9P:-Nu3yCT52iAxMAFqff1j:bxyewz1) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L26)):*

This is also consistent with the paper and the aptos repo.
</blockquote></details>

Got it. If Aptos initialized it to zero I guess it's possible to do so. However, I think it's more natural to initialize it to `n+1`:
1. At the beginning, you have no validation tasks available, so it's wrong to state that the validation index is zero.
2. For simplicity, let's assume we have one thread: the first tx is executed, and then `finish_execution` is called, which calls the `decrease_validation_idx` or returns a validation task. So, it adjusts the validation index counter as needed (decrease it to zero in this case).
3. If we have 2 threads, and the first caught the execution task for the first tx, the second might call `next_task`, there it will find that the validation index is zero whereas the execution index is one, so it calls `next_version_to_validate`, but it will not necessarily find `status = EXECUTED` so it will return None, and we got a sequence of redundant calls. (On the other hand, if we initialize the validation index to `n+1` the index will be decreased only after we have some tx with an `EXECUTED` status). 

(maybe I have some bugs in the description of the last point, but I hope the general idea is clear)

From the chainlink [blogpost](https://blog.chain.link/block-stm/):

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 103 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2rXFV8VM5AZ_BTaHl:-Nu42FNh16u5qpfMwvYD:bqk7gvx) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L103)):*

1. `self.validation_idx` does not work for reading the atomic variable's value. You have to `load` it via an atomic method.
2. We want the current `validation_idx`, not the next one. It is not very explicit in the paper, but it makes sense (e.g., if we always take the next one, we never validate/execute tx 0), and this is what they do in the aptos repo.

You are right some of the logic here is unnecessary, though, and I fixed the comments as you suggested.
</blockquote></details>

Nice, can you reference the part in the paper where you could understand it?
Notice that we must increment the `num_Active_tasks` before incrementing the `validation_index`. (this discussion is resolved, added a comment in the next revision).
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 78 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu41Xm3B73Yjh18KMvS:-Nu41Xm3B73Yjh18KMvT:b-6f26nb) ([raw file](https://github.com/starkware-libs/blockifier/blob/430a26dae99177070519124694dec8d10ff0b4eb/crates/blockifier/src/concurrency/scheduler.rs#L78)):*
> ```XML
> 
>     pub fn next_version_to_validate(&self) -> Option<u32> {
>         let index_to_validate = self.validation_index.fetch_add(1, Ordering::SeqCst);
> ```

We must increment the `num_Active_tasks` before incrementing the `validation_index`.
(move the `fetch_add` below)



_Suggestion:_
```XML
let index_to_validate = self.validation_index.load(Ordering::Acquire);
```
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 89 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu43Cdo2ndCTgHmr0lK:-Nu43CdpABxXcem0E182:b2nreh6) ([raw file](https://github.com/starkware-libs/blockifier/blob/430a26dae99177070519124694dec8d10ff0b4eb/crates/blockifier/src/concurrency/scheduler.rs#L89)):*
> ```XML
> 
>     pub fn next_version_to_execute(&self) -> Option<u32> {
>         let index_to_execute = self.execution_index.fetch_add(1, Ordering::SeqCst);
> ```

Same as above.



_Suggestion:_
```XML
let index_to_execute = self.execution_index.load(Ordering::Acquire);
```

noaov1 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1702)* status: all files reviewed, 10 unresolved discussions (waiting on @avi-starkware and @barak-b-starkware)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 26 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2l1oo9iiqPSuXIu9P:-Nu3zbI30FNoChN8LtQf:by5fkys) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L26)):*

This is also consistent with the paper and the aptos repo.
</blockquote></details>

Note- we start by executing the first transaction and than, trying to validate it (if it's ready). Otherwise we continue executing and so forth. We start validating as soon as the first transaction is ready to be validated.
If we will set the validation index to `chunk_size + 1` (why `+1`?), than we will start validating the first transaction that finished execution (which might not be the first one, and can lead to redundant validations).
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 85 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2qGpj1wNlzquSEtV9:-Nu42fFf5-eeSu5dOS_b:b-3nvrul) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L85)):*

I don't mind either way regarding the name of the function. @noaov1 What do you think?

I made the wording of the doc slightly different than you suggested.
</blockquote></details>

I find the word "incarnate" not very informative, but whatever you prefer :) 
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 100 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu42HCi0K_2jGLqSYb-:-Nu42HCi0K_2jGLqSYb0:boonp6f) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L100)):*
> ```XML
>             // if tx_status == TxStatus::Executed
>             return Some(idx_to_validate);
>         }
> ```

Why did you remove this `if`?



_Code quote:_
```XML
        if idx_to_validate < self.chunk_size {
            // TODO(barak, 01/04/2024): implement tx_status and add the following condition:
            // if tx_status == TxStatus::Executed
            return Some(idx_to_validate);
        }
```
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 36 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu40gCy5jIWcr-8cOMI:-Nu40gCy5jIWcr-8cOMJ:b1nf0us) ([raw file](https://github.com/starkware-libs/blockifier/blob/430a26dae99177070519124694dec8d10ff0b4eb/crates/blockifier/src/concurrency/scheduler.rs#L36)):*
> ```XML
>     }
> 
>     /// Returns the next task to run. Prioritizes execution tasks.
> ```

Why `execution`?



_Code quote:_
```XML
/// Returns the next task to run. Prioritizes execution tasks.
```
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 82 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu43pnY19xkdHNbOPCy:-Nu43pnY19xkdHNbOPCz:b-6sklfd) ([raw file](https://github.com/starkware-libs/blockifier/blob/430a26dae99177070519124694dec8d10ff0b4eb/crates/blockifier/src/concurrency/scheduler.rs#L82)):*
> ```XML
>             self.check_done();
>             return None;
>         }
> ```

Is it possible to have an overflow? We keep incrementing the index until we are done. 



_Code quote:_
```XML
        let index_to_validate = self.validation_index.fetch_add(1, Ordering::SeqCst);
        if index_to_validate >= self.chunk_size {
            self.check_done();
            return None;
        }
```

noaov1 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1702)* status: 1 of 3 files reviewed, 10 unresolved discussions (waiting on @avi-starkware and @barak-b-starkware)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 36 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu40gCy5jIWcr-8cOMI:-Nu49sAI4wSDfj-tnCDT:b9g56f7) ([raw file](https://github.com/starkware-libs/blockifier/blob/430a26dae99177070519124694dec8d10ff0b4eb/crates/blockifier/src/concurrency/scheduler.rs#L36)):*

Because it chooses the validation task only if `validation_index < execution_index`. If both equal `3` then it will return an execution task.
</blockquote></details>

Suggestion:
Prioritize execution and validation tasks for lower transactions.

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 26 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2l1oo9iiqPSuXIu9P:-Nu4Edp-ENB8Vu55aZbt:b8q84l7) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L26)):*

Note- we start by executing the first transaction and than, trying to validate it (if it's ready). Otherwise we continue executing and so forth. We start validating as soon as the first transaction is ready to be validated.
If we will set the validation index to `chunk_size + 1` (why `+1`?), than we will start validating the first transaction that finished execution (which might not be the first one, and can lead to redundant validations).
</blockquote></details>

I think this optimization saves us a small number of redundant steps and is not very intuitive, so I believe we should keep it initialized at 0.

avi-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1702)* status: 1 of 3 files reviewed, 10 unresolved discussions (waiting on @barak-b-starkware and @noaov1)

___
*[`-- commits` line 2 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2uY82-68GYX4WQeLc):*

Do we want to add new scope `concurrency` in addition to `execution` in the commitlint configuration? so it would be something like:
`feat(concurrency, execution): add counters to the scheduler`
@noaov1
</blockquote></details>

Done.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 85 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2qGpj1wNlzquSEtV9:-Nu4Cmf25p0WVSvnTQnb:becjl98) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L85)):*

I find the word "incarnate" not very informative, but whatever you prefer :)
</blockquote></details>

How about `try_tx_status_executing`?
`try_incarnate` is nice because it is the same as the paper
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 100 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu42HCi0K_2jGLqSYb-:-Nu4E2h81zxSLX-c2daJ:b-smr5g5) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L100)):*

Why did you remove this `if`?
</blockquote></details>

I reverted the change anyway.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 36 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu40gCy5jIWcr-8cOMI:-Nu4FNjBB6p5xb1N4Zcu:bicg46c) ([raw file](https://github.com/starkware-libs/blockifier/blob/430a26dae99177070519124694dec8d10ff0b4eb/crates/blockifier/src/concurrency/scheduler.rs#L36)):*

Suggestion:
Prioritize execution and validation tasks for lower transactions.
</blockquote></details>

Maybe we should just delete this sentence. The condition itself `validation_index < execution_index` is much clearer than the explanation.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 78 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu41Xm3B73Yjh18KMvS:-Nu48Icx6jqqe3PqM1T_:b-2j9vlq) ([raw file](https://github.com/starkware-libs/blockifier/blob/430a26dae99177070519124694dec8d10ff0b4eb/crates/blockifier/src/concurrency/scheduler.rs#L78)):*

We must increment the `num_Active_tasks` before incrementing the `validation_index`.
(move the `fetch_add` below)
</blockquote></details>

Good catch!
This is a very subtle bug. Another thread might get to `check_done` before the n_active_tasks gets incremented, and the done marker might become true even though there is an active task running!
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 82 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu43pnY19xkdHNbOPCy:-Nu4Dc2o-opE38GfLNFZ:b-agvlqn) ([raw file](https://github.com/starkware-libs/blockifier/blob/430a26dae99177070519124694dec8d10ff0b4eb/crates/blockifier/src/concurrency/scheduler.rs#L82)):*

Is it possible to have an overflow? We keep incrementing the index until we are done.
</blockquote></details>

Overflow does not matter. validation_index == chunk_size is the same as validation_index > chunk_size as far as the algorithm is concerned.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 89 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu43Cdo2ndCTgHmr0lK:-Nu4EEhADUH6KATJy2iS:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/430a26dae99177070519124694dec8d10ff0b4eb/crates/blockifier/src/concurrency/scheduler.rs#L89)):*

Same as above.
</blockquote></details>

Done.

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 103 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2rXFV8VM5AZ_BTaHl:-Nu4JZKz2Ed9I85gxxKR:b7y2s31) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L103)):*
See page 10, subsubsection titled "Managing Tasks", second paragraph:
> The validation_idx counter tracks the index of the *next transaction to be validated*.

As I said, it is not very explicit.

barak-b-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 36 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu40gCy5jIWcr-8cOMI:-Nu4PHRWArwTfo9LOiLx:b-8ayu6x) ([raw file](https://github.com/starkware-libs/blockifier/blob/430a26dae99177070519124694dec8d10ff0b4eb/crates/blockifier/src/concurrency/scheduler.rs#L36)):*

Maybe we should just delete this sentence. The condition itself `validation_index < execution_index` is much clearer than the explanation.
</blockquote></details>

Agree

barak-b-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 85 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2qGpj1wNlzquSEtV9:-Nu4PNOl92TRnMxmDuTd:bgq1ayl) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L85)):*

How about `try_tx_status_executing`?
`try_incarnate` is nice because it is the same as the paper
</blockquote></details>

I vote for `try_incarnate` since it's the same as the paper, and I resolve this issue for now (:

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 36 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu40gCy5jIWcr-8cOMI:-Nu4PtRT8F-VUI5Yr6Uv:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/430a26dae99177070519124694dec8d10ff0b4eb/crates/blockifier/src/concurrency/scheduler.rs#L36)):*

Agree
</blockquote></details>

Done.

barak-b-starkware (changes requested):
> Reviewed 2 of 2 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1702)* status: all files reviewed, 8 unresolved discussions (waiting on @avi-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 26 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2l1oo9iiqPSuXIu9P:-Nu4QxDa7GO5rGIbCfpK:big57qy) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L26)):*

I think this optimization saves us a small number of redundant steps and is not very intuitive, so I believe we should keep it initialized at 0.
</blockquote></details>

After discussing with @noaov1 offline, I think my main point is that the nature of the algorithm implies initializing it to `n+1` (or `n`). The optimization here is initializing it to zero, which gives almost nothing. Moreover, it's just a mid-stage in the algorithm evolution (to avoid optimizations and keep it as simple as possible), so making an optimization for a mid-stage that should not be optimized is weird.
Overall, I vote for the `n` choice since (AFAIU) it's the "right" option algorithm-wise, and the zero choice is the optimized one.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 103 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2rXFV8VM5AZ_BTaHl:-Nu4Pz56CMS7jknXdXVu:bwfpavu) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L103)):*

See page 10, subsubsection titled "Managing Tasks", second paragraph:
> The validation_idx counter tracks the index of the *next transaction to be validated*.

As I said, it is not very explicit.
</blockquote></details>

Nice thanks!
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 91 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu4QJEK7ttHw1B1VQOE:-Nu4QJEK7ttHw1B1VQOF:b-uwm7hn) ([raw file](https://github.com/starkware-libs/blockifier/blob/d4f14ce2fa49ab0133eb2602b15c4f45de4b79a8/crates/blockifier/src/concurrency/scheduler.rs#L91)):*
> ```XML
>         self.n_active_tasks.fetch_sub(1, Ordering::SeqCst);
>         None
>     }
> ```

Will this work?



_Suggestion:_
```XML
    pub fn next_version_to_validate(&self) -> Option<u32> {
        let index_to_validate = self.validation_index.load(Ordering::Acquire);
        if index_to_validate >= self.chunk_size {
            self.check_done();
            return None;
        }
        self.n_active_tasks.fetch_add(1, Ordering::SeqCst);
        self.validation_index.fetch_add(1, Ordering::SeqCst);
        if index_to_validate < self.chunk_size {
            // TODO(barak, 01/04/2024): complete next_version_to_validate logic.
            return Some(index_to_validate);
        }
        self.n_active_tasks.fetch_sub(1, Ordering::SeqCst);
        None
    }
```
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 102 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu4Q_a6EQRSB0-98RJN:-Nu4Q_a7CfDyCg0o5HwJ:b-uwm7hn) ([raw file](https://github.com/starkware-libs/blockifier/blob/d4f14ce2fa49ab0133eb2602b15c4f45de4b79a8/crates/blockifier/src/concurrency/scheduler.rs#L102)):*
> ```XML
>         index_to_execute = self.execution_index.fetch_add(1, Ordering::SeqCst);
>         self.try_incarnate(index_to_execute)
>     }
> ```

Will this work?



_Suggestion:_
```XML
    pub fn next_version_to_execute(&self) -> Option<u32> {
        let index_to_execute = self.execution_index.load(Ordering::Acquire);
        if index_to_execute >= self.chunk_size {
            self.check_done();
            return None;
        }
        self.n_active_tasks.fetch_add(1, Ordering::SeqCst);
        self.execution_index.fetch_add(1, Ordering::SeqCst);
        self.try_incarnate(index_to_execute)
    }
```

barak-b-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 36 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu40gCy5jIWcr-8cOMI:-Nu4Sr2U7rirJbfizuVu:bc77tv0) ([raw file](https://github.com/starkware-libs/blockifier/blob/430a26dae99177070519124694dec8d10ff0b4eb/crates/blockifier/src/concurrency/scheduler.rs#L36)):*

Done.
</blockquote></details>

I think that the `/// Returns the next task to run.` is also redundant, it's just repeating "next_task"

barak-b-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 26 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2l1oo9iiqPSuXIu9P:-Nu4Tfh28h_hDWVSPiQH:b-wlecss) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L26)):*

After discussing with @noaov1 offline, I think my main point is that the nature of the algorithm implies initializing it to `n+1` (or `n`). The optimization here is initializing it to zero, which gives almost nothing. Moreover, it's just a mid-stage in the algorithm evolution (to avoid optimizations and keep it as simple as possible), so making an optimization for a mid-stage that should not be optimized is weird.
Overall, I vote for the `n` choice since (AFAIU) it's the "right" option algorithm-wise, and the zero choice is the optimized one.
</blockquote></details>

One last point: the full algorithm will not enjoy this optimization, so if we agree that the "right" option is `n` and we still choose to optimize it for now, we will need to change it to `n`.

noaov1 (commented):
> Reviewed 1 of 2 files at r6, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1702)* status: all files reviewed, 5 unresolved discussions (waiting on @avi-starkware)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 60 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu4Rv45-ijrvJzBXZGX:-Nu4Rv45-ijrvJzBXZGY:bjf8c5c) ([raw file](https://github.com/starkware-libs/blockifier/blob/80cb51222ef823f93d6f7329fe7f83d5e6942cb4/crates/blockifier/src/concurrency/scheduler.rs#L60)):*
> ```XML
>         self.validation_index.fetch_min(target_idx, Ordering::SeqCst);
>         self.decrease_counter.fetch_add(1, Ordering::SeqCst);
>     }
> ```

Same for `decrease_execution_index` and `check_done`.



_Suggestion:_
```XML
    fn decrease_validation_index(&self, target_idx: u32) {
        self.validation_index.fetch_min(target_idx, Ordering::SeqCst);
        self.decrease_counter.fetch_add(1, Ordering::SeqCst);
    }
```
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 84 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu4U3W7BJB60qXdAmWZ:-Nu4U3W7BJB60qXdAmW_:b45h) ([raw file](https://github.com/starkware-libs/blockifier/blob/80cb51222ef823f93d6f7329fe7f83d5e6942cb4/crates/blockifier/src/concurrency/scheduler.rs#L84)):*
> ```XML
>         }
>         self.n_active_tasks.fetch_add(1, Ordering::SeqCst);
>         index_to_validate = self.validation_index.fetch_add(1, Ordering::SeqCst);
> ```





_Suggestion:_
```XML
        let index_to_validate = self.validation_index.load(Ordering::Acquire);
        if index_to_validate >= self.chunk_size {
            self.check_done();
            return None;
        }
        self.n_active_tasks.fetch_add(1, Ordering::SeqCst);
        let index_to_validate = self.validation_index.fetch_add(1, Ordering::SeqCst);
```

barak-b-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 84 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu4U3W7BJB60qXdAmWZ:-Nu4V4wkAz4ztacbhen6:bug557w) ([raw file](https://github.com/starkware-libs/blockifier/blob/80cb51222ef823f93d6f7329fe7f83d5e6942cb4/crates/blockifier/src/concurrency/scheduler.rs#L84)):*
> ```XML
>         }
>         self.n_active_tasks.fetch_add(1, Ordering::SeqCst);
>         index_to_validate = self.validation_index.fetch_add(1, Ordering::SeqCst);
> ```

Why create a new variable and not use the one from the `load`? see [here](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu4QJEK7ttHw1B1VQOE)

barak-b-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 60 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu4Rv45-ijrvJzBXZGX:-Nu4VxJlBHaWwdkaWNOI:b-z8lc8z) ([raw file](https://github.com/starkware-libs/blockifier/blob/80cb51222ef823f93d6f7329fe7f83d5e6942cb4/crates/blockifier/src/concurrency/scheduler.rs#L60)):*

Same for `decrease_execution_index` and `check_done`.
</blockquote></details>

Nice! and same for `try_incarnate`, `next_version_to_validate`, `next_version_to_execute`.

noaov1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1702)* status: all files reviewed, 5 unresolved discussions (waiting on @avi-starkware and @barak-b-starkware)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 84 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu4U3W7BJB60qXdAmWZ:-Nu4Vv2f77RpU5YlC5gI:b-5rx1v1) ([raw file](https://github.com/starkware-libs/blockifier/blob/80cb51222ef823f93d6f7329fe7f83d5e6942cb4/crates/blockifier/src/concurrency/scheduler.rs#L84)):*

Why create a new variable and not use the one from the `load`? see [here](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu4QJEK7ttHw1B1VQOE)
</blockquote></details>

Then you need to use `mut`. It is technically not the same variable, but have the same name.

barak-b-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 84 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu4U3W7BJB60qXdAmWZ:-Nu4YThN4iJPJExDOo2P:b-ccmmxf) ([raw file](https://github.com/starkware-libs/blockifier/blob/80cb51222ef823f93d6f7329fe7f83d5e6942cb4/crates/blockifier/src/concurrency/scheduler.rs#L84)):*

Then you need to use `mut`. It is technically not the same variable, but have the same name.
</blockquote></details>

Not sure I follow.
1. you load `validation_index` to `index_to_validate` (`index_to_validate` shouldn't be changed from now and can be returned as is).
2. you increment `validation_index` after incrementing `n_active_tasks`.
3. if ... you return `index_to_validate` that was loaded initially.

(Notice that `fetch_add` returns the value before incrementing).
What I miss here?

noaov1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1702)* status: all files reviewed, 5 unresolved discussions (waiting on @avi-starkware and @barak-b-starkware)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 84 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu4U3W7BJB60qXdAmWZ:-Nu4d3fa14GcoZrOSOKz:b-rh60d8) ([raw file](https://github.com/starkware-libs/blockifier/blob/80cb51222ef823f93d6f7329fe7f83d5e6942cb4/crates/blockifier/src/concurrency/scheduler.rs#L84)):*

Not sure I follow.
1. you load `validation_index` to `index_to_validate` (`index_to_validate` shouldn't be changed from now and can be returned as is).
2. you increment `validation_index` after incrementing `n_active_tasks`.
3. if ... you return `index_to_validate` that was loaded initially.

(Notice that `fetch_add` returns the value before incrementing).
What I miss here?
</blockquote></details>

Oh, I see what you meant. AFAIU, you need to fetch again, in case the validation index was changed in between (by another thread).

avi-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1702)* status: all files reviewed, 4 unresolved discussions (waiting on @barak-b-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 26 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2l1oo9iiqPSuXIu9P:-Nu53F4d2T9pcEFYqhKO:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L26)):*

One last point: the full algorithm will not enjoy this optimization, so if we agree that the "right" option is `n` and we still choose to optimize it for now, we will need to change it to `n`.
</blockquote></details>

Done.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 85 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu2qGpj1wNlzquSEtV9:-Nu53kvFBzupGlbgIdA6:b-837i3v) ([raw file](https://github.com/starkware-libs/blockifier/blob/d53d2121c6d5157aea9f0a97598a0c903863fd35/crates/blockifier/src/concurrency/scheduler.rs#L85)):*

I vote for `try_incarnate` since it's the same as the paper, and I resolve this issue for now (:
</blockquote></details>

Okay.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 36 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu40gCy5jIWcr-8cOMI:-Nu530aZ4Q5aufcLq7fq:bss4a2b) ([raw file](https://github.com/starkware-libs/blockifier/blob/430a26dae99177070519124694dec8d10ff0b4eb/crates/blockifier/src/concurrency/scheduler.rs#L36)):*

I think that the `/// Returns the next task to run.` is also redundant, it's just repeating "next_task"
</blockquote></details>

Removed.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 91 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu4QJEK7ttHw1B1VQOE:-Nu54YdS0mRSkx9HvJLH:b-2cr2mw) ([raw file](https://github.com/starkware-libs/blockifier/blob/d4f14ce2fa49ab0133eb2602b15c4f45de4b79a8/crates/blockifier/src/concurrency/scheduler.rs#L91)):*

Will this work?
</blockquote></details>

This will not work because two threads can get the same validation index.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 102 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu4Q_a6EQRSB0-98RJN:-Nu54k8tEq1euEfzn07p:b-q64pkt) ([raw file](https://github.com/starkware-libs/blockifier/blob/d4f14ce2fa49ab0133eb2602b15c4f45de4b79a8/crates/blockifier/src/concurrency/scheduler.rs#L102)):*

Will this work?
</blockquote></details>

See above.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 60 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu4Rv45-ijrvJzBXZGX:-Nu53Yjv8VPwvaLCoq_t:bs5nuhn) ([raw file](https://github.com/starkware-libs/blockifier/blob/80cb51222ef823f93d6f7329fe7f83d5e6942cb4/crates/blockifier/src/concurrency/scheduler.rs#L60)):*

Nice! and same for `try_incarnate`, `next_version_to_validate`, `next_version_to_execute`.
</blockquote></details>

Done.

I made all the fields private as well.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 84 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu4U3W7BJB60qXdAmWZ:-Nu53wvs5PtjVRyj-2I-:b-qhbs1y) ([raw file](https://github.com/starkware-libs/blockifier/blob/80cb51222ef823f93d6f7329fe7f83d5e6942cb4/crates/blockifier/src/concurrency/scheduler.rs#L84)):*

Oh, I see what you meant. AFAIU, you need to fetch again, in case the validation index was changed in between (by another thread).
</blockquote></details>

Another thread could have snatched this validation index before you incremented.
What is the problem with using `mut`?
Reverted to non-`mut` for now.

noaov1 (approved):
> Reviewed 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1702)* status: all files reviewed, 5 unresolved discussions (waiting on @avi-starkware and @barak-b-starkware)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 84 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu4U3W7BJB60qXdAmWZ:-NuI6u8o8eIfAfQaioEA:bpu3gk7) ([raw file](https://github.com/starkware-libs/blockifier/blob/80cb51222ef823f93d6f7329fe7f83d5e6942cb4/crates/blockifier/src/concurrency/scheduler.rs#L84)):*

Another thread could have snatched this validation index before you incremented.
What is the problem with using `mut`?
Reverted to non-`mut` for now.
</blockquote></details>

You don't really change the variable, but simply use the same name twice. It's better to avoid `mut`.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 27 at r8](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-NuI6CzR62sK7adnMLPl:-NuI6CzR62sK7adnMLPm:undefined) ([raw file](https://github.com/starkware-libs/blockifier/blob/3ee92ea4391b265c059b0bcf6da3d6f69af69f54/crates/blockifier/src/concurrency/scheduler.rs#L27)):*
> ```XML
> impl Scheduler {
>     pub fn new(chunk_size: u32) -> Scheduler {
>         Scheduler { chunk_size, validation_index: (chunk_size + 1).into(), ..Default::default() }
> ```





_Suggestion:_
```XML
        Scheduler { chunk_size, validation_index: chunk_size.into(), ..Default::default() }
```

barak-b-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 84 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu4U3W7BJB60qXdAmWZ:-Nu5I5Ze-dpeX_sj1zAB:bwfomyb) ([raw file](https://github.com/starkware-libs/blockifier/blob/80cb51222ef823f93d6f7329fe7f83d5e6942cb4/crates/blockifier/src/concurrency/scheduler.rs#L84)):*

You don't really change the variable, but simply use the same name twice. It's better to avoid `mut`.
</blockquote></details>

Nice, right, it can be tricky, and my first suggestion is wrong.

So. as you say, potentially, you do change the variable, otherwise, you could just load it in the beginning, if I'm not wrong, I think the `mut` option is better.

barak-b-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 91 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu4QJEK7ttHw1B1VQOE:-NuIB5pI2RxHss-p8nOt:bddnk2m) ([raw file](https://github.com/starkware-libs/blockifier/blob/d4f14ce2fa49ab0133eb2602b15c4f45de4b79a8/crates/blockifier/src/concurrency/scheduler.rs#L91)):*

This will not work because two threads can get the same validation index.
</blockquote></details>

Right, thanks (:

barak-b-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 84 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-Nu4U3W7BJB60qXdAmWZ:-NuIEkyAE6qs1Bv1TEe7:bqnnmmi) ([raw file](https://github.com/starkware-libs/blockifier/blob/80cb51222ef823f93d6f7329fe7f83d5e6942cb4/crates/blockifier/src/concurrency/scheduler.rs#L84)):*

Nice, right, it can be tricky, and my first suggestion is wrong.

So. as you say, potentially, you do change the variable, otherwise, you could just load it in the beginning, if I'm not wrong, I think the `mut` option is better.
</blockquote></details>

OK, after an offline discussion with @noaov1, I think it's better to rebind it with a new `let` expression. The function doesn't mutate this variable, it (potentially) happens "behind the scenes" by another thread.

noaov1 (dismissed):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1702)* status: all files reviewed, 3 unresolved discussions (waiting on @avi-starkware and @barak-b-starkware)

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 27 at r8](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-NuI6CzR62sK7adnMLPl:-NuII7Br-y7R5DOOSgVQ:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/3ee92ea4391b265c059b0bcf6da3d6f69af69f54/crates/blockifier/src/concurrency/scheduler.rs#L27)):*
> ```XML
> impl Scheduler {
>     pub fn new(chunk_size: u32) -> Scheduler {
>         Scheduler { chunk_size, validation_index: (chunk_size + 1).into(), ..Default::default() }
> ```

Done.

noaov1 (approved):
:lgtm:

> Reviewed 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-:-NuIJ0Nw1bs74soGAoA-:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @barak-b-starkware)

barak-b-starkware (approved):
:lgtm:

> Reviewed 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1702#-:-NuIS_IB2PScV3Zcbkjb:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @avi-starkware)

PR #1795, opened 2024-04-17
feat(concurrency): use tx statuses in the scheduler

avi-starkware:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1795)

### DIFF
--- a/crates/blockifier/src/concurrency/scheduler.rs
+++ b/crates/blockifier/src/concurrency/scheduler.rs
@@ -1,6 +1,6 @@
 use std::cmp::min;
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
-use std::sync::Mutex;
+use std::sync::{Mutex, MutexGuard};
 
 use crate::concurrency::TxIndex;
 
@@ -20,6 +20,7 @@ pub struct Scheduler {
     decrease_counter: AtomicUsize,
     n_active_tasks: AtomicUsize,
     chunk_size: usize,
+    // TODO(Avi, 15/05/2024): Consider using RwLock instead of Mutex.
     tx_statuses: Box<[Mutex<TransactionStatus>]>,
     /// Updated by the `check_done` procedure, providing a cheap way for all threads to exit their
     /// main loops.
@@ -66,6 +67,16 @@ impl Scheduler {
     fn safe_decrement_n_active_tasks(&self) {
         let previous_n_active_tasks = self.n_active_tasks.fetch_sub(1, Ordering::SeqCst);
         assert!(previous_n_active_tasks > 0, "n_active_tasks underflow");
+    }
+
+    fn lock_tx_status(&self, tx_index: TxIndex) -> MutexGuard<'_, TransactionStatus> {
+        self.tx_statuses[tx_index].lock().unwrap_or_else(|error| {
+            panic!(
+                "Status of transaction index {} is poisoned. Data: {:?}.",
+                tx_index,
+                *error.get_ref()
+            )
+        })
     }
 
     pub fn next_task(&self) -> Task {
@@ -128,8 +139,11 @@ impl Scheduler {
     /// Updates a transaction's status to `Executing` if it is ready to execute.
     fn try_incarnate(&self, tx_index: TxIndex) -> Option<TxIndex> {
         if tx_index < self.chunk_size {
-            // TODO(barak, 01/04/2024): complete try_incarnate logic.
-            return Some(tx_index);
+            let mut status = self.lock_tx_status(tx_index);
+            if *status == TransactionStatus::ReadyToExecute {
+                *status = TransactionStatus::Executing;
+                return Some(tx_index);
+            }
         }
         self.safe_decrement_n_active_tasks();
         None
@@ -144,8 +158,10 @@ impl Scheduler {
         self.n_active_tasks.fetch_add(1, Ordering::SeqCst);
         let index_to_validate = self.validation_index.fetch_add(1, Ordering::SeqCst);
         if index_to_validate < self.chunk_size {
-            // TODO(barak, 01/04/2024): complete next_version_to_validate logic.
-            return Some(index_to_validate);
+            let status = self.lock_tx_status(index_to_validate);
+            if *status == TransactionStatus::Executed {
+                return Some(index_to_validate);
+            }
         }
         self.safe_decrement_n_active_tasks();
         None
@@ -160,6 +176,14 @@ impl Scheduler {
         self.n_active_tasks.fetch_add(1, Ordering::SeqCst);
         let index_to_execute = self.execution_index.fetch_add(1, Ordering::SeqCst);
         self.try_incarnate(index_to_execute)
+    }
+
+    #[cfg(test)]
+    fn set_tx_status(&self, tx_index: TxIndex, status: TransactionStatus) {
+        if tx_index < self.chunk_size {
+            let mut tx_status = self.lock_tx_status(tx_index);
+            *tx_status = status;
+        }
     }
 }
 
--- a/crates/blockifier/src/concurrency/scheduler_test.rs
+++ b/crates/blockifier/src/concurrency/scheduler_test.rs
@@ -1,5 +1,6 @@
 use std::cmp::min;
 use std::sync::atomic::Ordering;
+use std::sync::Arc;
 
 use pretty_assertions::assert_eq;
 use rstest::rstest;
@@ -31,8 +32,8 @@ fn test_new(#[values(0, 1, 32)] chunk_size: usize) {
 #[case::execution_incomplete(DEFAULT_CHUNK_SIZE-1, DEFAULT_CHUNK_SIZE+1, 0, false)]
 #[case::validation_incomplete(DEFAULT_CHUNK_SIZE, DEFAULT_CHUNK_SIZE-1, 0, false)]
 fn test_check_done(
-    #[case] execution_index: usize,
-    #[case] validation_index: usize,
+    #[case] execution_index: TxIndex,
+    #[case] validation_index: TxIndex,
     #[case] n_active_tasks: usize,
     #[case] expected: bool,
 ) {
@@ -47,119 +48,174 @@ fn test_check_done(
 }
 
 #[rstest]
-#[case::happy_flow(1, 0)]
+#[case::no_panic(1)]
 #[should_panic(expected = "n_active_tasks underflow")]
-#[case::underflow(0, 0)]
-fn test_safe_decrement_n_active_tasks(
-    #[case] n_active_tasks: usize,
-    #[case] expected_n_active_tasks: usize,
-) {
+#[case::underflow_panic(0)]
+fn test_safe_decrement_n_active_tasks(#[case] n_active_tasks: usize) {
     let scheduler =
         default_scheduler!(chunk_size: DEFAULT_CHUNK_SIZE, n_active_tasks: n_active_tasks);
     scheduler.safe_decrement_n_active_tasks();
-    assert_eq!(scheduler.n_active_tasks.load(Ordering::Acquire), expected_n_active_tasks);
+    assert_eq!(scheduler.n_active_tasks.load(Ordering::Acquire), n_active_tasks - 1);
 }
 
 #[rstest]
-#[case::done(DEFAULT_CHUNK_SIZE, DEFAULT_CHUNK_SIZE, true, Task::Done, 0)]
-#[case::no_task(DEFAULT_CHUNK_SIZE, DEFAULT_CHUNK_SIZE, false, Task::NoTask, 0)]
-#[case::execution_task(0, 0, false, Task::ExecutionTask(0), 1)]
-#[case::validation_task(1, 0, false, Task::ValidationTask(0), 1)]
+fn test_lock_tx_status() {
+    let scheduler = Scheduler::new(DEFAULT_CHUNK_SIZE);
+    let status = scheduler.lock_tx_status(0);
+    assert_eq!(*status, TransactionStatus::ReadyToExecute);
+}
+
+#[rstest]
+#[should_panic(expected = "Status of transaction index 0 is poisoned. Data: ReadyToExecute.")]
+fn test_lock_tx_status_poisoned() {
+    let scheduler = Arc::new(Scheduler::new(DEFAULT_CHUNK_SIZE));
+    let scheduler_clone = scheduler.clone();
+    let handle = std::thread::spawn(move || {
+        let _guard = scheduler_clone.lock_tx_status(0);
+        panic!("Intentional panic to poison the mutex")
+    });
+    handle.join().expect_err("Thread did not panic as expected");
+    // The panic is expected here.
+    let _guard = scheduler.lock_tx_status(0);
+}
+
+#[rstest]
+#[case::done(DEFAULT_CHUNK_SIZE, DEFAULT_CHUNK_SIZE, TransactionStatus::Executed, Task::Done)]
+#[case::no_task(DEFAULT_CHUNK_SIZE, DEFAULT_CHUNK_SIZE, TransactionStatus::Executed, Task::NoTask)]
+#[case::no_task_as_validation_index_not_executed(
+    DEFAULT_CHUNK_SIZE,
+    0,
+    TransactionStatus::ReadyToExecute,
+    Task::NoTask
+)]
+#[case::execution_task(0, 0, TransactionStatus::ReadyToExecute, Task::ExecutionTask(0))]
+#[case::execution_task_as_validation_index_not_executed(
+    1,
+    0,
+    TransactionStatus::ReadyToExecute,
+    Task::ExecutionTask(1)
+)]
+#[case::validation_task(1, 0, TransactionStatus::Executed, Task::ValidationTask(0))]
 fn test_next_task(
-    #[case] execution_index: usize,
-    #[case] validation_index: usize,
-    #[case] done_marker: bool,
+    #[case] execution_index: TxIndex,
+    #[case] validation_index: TxIndex,
+    #[case] validation_index_status: TransactionStatus,
     #[case] expected_next_task: Task,
-    #[case] expected_n_active_tasks: usize,
 ) {
     let scheduler = default_scheduler!(
         chunk_size: DEFAULT_CHUNK_SIZE,
         execution_index: execution_index,
         validation_index: validation_index,
-        done_marker: done_marker,
+        done_marker: expected_next_task == Task::Done,
     );
+    scheduler.set_tx_status(validation_index, validation_index_status);
     let next_task = scheduler.next_task();
     assert_eq!(next_task, expected_next_task);
+    let expected_n_active_tasks = match expected_next_task {
+        Task::Done | Task::NoTask => 0,
+        _ => 1,
+    };
     assert_eq!(scheduler.n_active_tasks.load(Ordering::Acquire), expected_n_active_tasks);
 }
 
 #[rstest]
-#[case::target_index_lt_validation_index(1, 3, 1)]
-#[case::target_index_eq_validation_index(3, 3, 0)]
-#[case::target_index_eq_validation_index_eq_zero(0, 0, 0)]
-#[case::target_index_gt_validation_index(1, 0, 0)]
+#[case::target_index_lt_validation_index(1, 3)]
+#[case::target_index_eq_validation_index(3, 3)]
+#[case::target_index_eq_validation_index_eq_zero(0, 0)]
+#[case::target_index_gt_validation_index(1, 0)]
 fn test_decrease_validation_index(
     #[case] target_index: TxIndex,
-    #[case] validation_index: usize,
-    #[case] expected_decrease_counter: usize,
+    #[case] validation_index: TxIndex,
 ) {
     let scheduler =
         default_scheduler!(chunk_size: DEFAULT_CHUNK_SIZE, validation_index: validation_index);
     scheduler.decrease_validation_index(target_index);
     let expected_validation_index = min(target_index, validation_index);
     assert_eq!(scheduler.validation_index.load(Ordering::Acquire), expected_validation_index);
+    let expected_decrease_counter = if target_index < validation_index { 1 } else { 0 };
     assert_eq!(scheduler.decrease_counter.load(Ordering::Acquire), expected_decrease_counter);
 }
 
 #[rstest]
-#[case::target_index_lt_execution_index(1, 3, 1)]
-#[case::target_index_eq_execution_index(3, 3, 0)]
-#[case::target_index_eq_execution_index_eq_zero(0, 0, 0)]
-#[case::target_index_gt_execution_index(1, 0, 0)]
-fn test_decrease_execution_index(
-    #[case] target_index: TxIndex,
-    #[case] execution_index: usize,
-    #[case] expected_decrease_counter: usize,
-) {
+#[case::target_index_lt_execution_index(1, 3)]
+#[case::target_index_eq_execution_index(3, 3)]
+#[case::target_index_eq_execution_index_eq_zero(0, 0)]
+#[case::target_index_gt_execution_index(1, 0)]
+fn test_decrease_execution_index(#[case] target_index: TxIndex, #[case] execution_index: TxIndex) {
     let scheduler =
         default_scheduler!(chunk_size: DEFAULT_CHUNK_SIZE, execution_index: execution_index);
     scheduler.decrease_execution_index(target_index);
     let expected_execution_index = min(target_index, execution_index);
     assert_eq!(scheduler.execution_index.load(Ordering::Acquire), expected_execution_index);
+    let expected_decrease_counter = if target_index < execution_index { 1 } else { 0 };
     assert_eq!(scheduler.decrease_counter.load(Ordering::Acquire), expected_decrease_counter);
 }
 
 #[rstest]
-#[case::from_ready_to_execute_to_executing(0, Some(0), 1)]
-#[case::index_out_of_bounds(DEFAULT_CHUNK_SIZE, None, 0)]
+#[case::ready_to_execute(0, TransactionStatus::ReadyToExecute, Some(0))]
+#[case::executing(0, TransactionStatus::Executing, None)]
+#[case::executed(0, TransactionStatus::Executed, None)]
+#[case::aborting(0, TransactionStatus::Aborting, None)]
+#[case::index_out_of_bounds(DEFAULT_CHUNK_SIZE, TransactionStatus::ReadyToExecute, None)]
 fn test_try_incarnate(
-    #[case] tx_index: usize,
-    #[case] expected_output: Option<usize>,
-    #[case] expected_n_active_tasks: usize,
+    #[case] tx_index: TxIndex,
+    #[case] tx_status: TransactionStatus,
+    #[case] expected_output: Option<TxIndex>,
 ) {
     let scheduler = default_scheduler!(chunk_size: DEFAULT_CHUNK_SIZE, n_active_tasks: 1);
+    scheduler.set_tx_status(tx_index, tx_status);
     assert_eq!(scheduler.try_incarnate(tx_index), expected_output);
-    assert_eq!(scheduler.n_active_tasks.load(Ordering::Acquire), expected_n_active_tasks);
+    if expected_output.is_some() {
+        assert_eq!(*scheduler.lock_tx_status(tx_index), TransactionStatus::Executing);
+        assert_eq!(scheduler.n_active_tasks.load(Ordering::Acquire), 1);
+    } else {
+        assert_eq!(scheduler.n_active_tasks.load(Ordering::Acquire), 0);
+        if tx_index < DEFAULT_CHUNK_SIZE {
+            assert_eq!(*scheduler.lock_tx_status(tx_index), tx_status);
+        }
+    }
 }
 
 #[rstest]
-#[case::some(1, Some(1), 2, 1)]
-#[case::none(DEFAULT_CHUNK_SIZE, None, DEFAULT_CHUNK_SIZE, 0)]
+#[case::ready_to_execute(1, TransactionStatus::ReadyToExecute, None)]
+#[case::executing(1, TransactionStatus::Executing, None)]
+#[case::executed(1, TransactionStatus::Executed, Some(1))]
+#[case::aborting(1, TransactionStatus::Aborting, None)]
+#[case::index_out_of_bounds(DEFAULT_CHUNK_SIZE, TransactionStatus::ReadyToExecute, None)]
 fn test_next_version_to_validate(
-    #[case] validation_index: usize,
-    #[case] expected_output: Option<usize>,
-    #[case] expected_validation_index: usize,
-    #[case] expected_n_active_tasks: usize,
+    #[case] validation_index: TxIndex,
+    #[case] tx_status: TransactionStatus,
+    #[case] expected_output: Option<TxIndex>,
 ) {
     let scheduler =
         default_scheduler!(chunk_size: DEFAULT_CHUNK_SIZE, validation_index: validation_index);
+    scheduler.set_tx_status(validation_index, tx_status);
     assert_eq!(scheduler.next_version_to_validate(), expected_output);
+    let expected_validation_index =
+        if validation_index < DEFAULT_CHUNK_SIZE { validation_index + 1 } else { validation_index };
     assert_eq!(scheduler.validation_index.load(Ordering::Acquire), expected_validation_index);
+    let expected_n_active_tasks = if expected_output.is_some() { 1 } else { 0 };
     assert_eq!(scheduler.n_active_tasks.load(Ordering::Acquire), expected_n_active_tasks);
 }
 
 #[rstest]
-#[case::some(1, Some(1), 2, 1)]
-#[case::none(DEFAULT_CHUNK_SIZE, None, DEFAULT_CHUNK_SIZE, 0)]
+#[case::ready_to_execute(1, TransactionStatus::ReadyToExecute, Some(1))]
+#[case::executing(1, TransactionStatus::Executing, None)]
+#[case::executed(1, TransactionStatus::Executed, None)]
+#[case::aborting(1, TransactionStatus::Aborting, None)]
+#[case::index_out_of_bounds(DEFAULT_CHUNK_SIZE, TransactionStatus::ReadyToExecute, None)]
 fn test_next_version_to_execute(
-    #[case] execution_index: usize,
-    #[case] expected_output: Option<usize>,
-    #[case] expected_execution_index: usize,
-    #[case] expected_n_active_tasks: usize,
+    #[case] execution_index: TxIndex,
+    #[case] tx_status: TransactionStatus,
+    #[case] expected_output: Option<TxIndex>,
 ) {
     let scheduler =
         default_scheduler!(chunk_size: DEFAULT_CHUNK_SIZE, execution_index: execution_index);
+    scheduler.set_tx_status(execution_index, tx_status);
     assert_eq!(scheduler.next_version_to_execute(), expected_output);
+    let expected_execution_index =
+        if execution_index < DEFAULT_CHUNK_SIZE { execution_index + 1 } else { execution_index };
     assert_eq!(scheduler.execution_index.load(Ordering::Acquire), expected_execution_index);
+    let expected_n_active_tasks = if expected_output.is_some() { 1 } else { 0 };
     assert_eq!(scheduler.n_active_tasks.load(Ordering::Acquire), expected_n_active_tasks);
 }


noaov1 (changes requested):
> Reviewed 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1795)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @avi-starkware and @barak-b-starkware)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 126 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NvgBAgb2Gk65alonu-X:-NvgBAgb2Gk65alonu-Y:bhix4ml) ([raw file](https://github.com/starkware-libs/blockifier/blob/4f50c8409dcdbc1ddb860cb3d4f02a0a683a5aa8/crates/blockifier/src/concurrency/scheduler.rs#L126)):*
> ```XML
>     fn try_incarnate(&self, tx_index: TxIndex) -> Option<TxIndex> {
>         if tx_index < self.chunk_size {
>             let mut status = self.tx_statuses[tx_index].try_lock().unwrap();
> ```

I think it panics only when called if the lock is already held by the current thread.



_Suggestion:_
```XML
let mut status = self.tx_statuses[tx_index].lock().expect("Transaction statuses is poisened.");
```
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 128 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nvk0ZL28OCpOMxakxWY:-Nvk0ZL28OCpOMxakxWZ:b9va5lh) ([raw file](https://github.com/starkware-libs/blockifier/blob/4f50c8409dcdbc1ddb860cb3d4f02a0a683a5aa8/crates/blockifier/src/concurrency/scheduler.rs#L128)):*
> ```XML
>             let mut status = self.tx_statuses[tx_index].try_lock().unwrap();
>             if *status == TransactionStatus::ReadyToExecute {
>                 *status = TransactionStatus::Executing;
> ```

Will you update the status in the `tx_statuses` array? Hoe does it work with the lock? You holds it until then?



_Code quote:_
```XML
                *status = TransactionStatus::Executing;
```

avi-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1795)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @barak-b-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 126 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NvgBAgb2Gk65alonu-X:-Nvkcwfg4RCNRnjZ7ZWC:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/4f50c8409dcdbc1ddb860cb3d4f02a0a683a5aa8/crates/blockifier/src/concurrency/scheduler.rs#L126)):*

I think it panics only when called if the lock is already held by the current thread.
</blockquote></details>

Done.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 128 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nvk0ZL28OCpOMxakxWY:-Nvkcx7u9LXda6aTALP3:bpo5has) ([raw file](https://github.com/starkware-libs/blockifier/blob/4f50c8409dcdbc1ddb860cb3d4f02a0a683a5aa8/crates/blockifier/src/concurrency/scheduler.rs#L128)):*

Will you update the status in the `tx_statuses` array? Hoe does it work with the lock? You holds it until then?
</blockquote></details>

The lock is held until the end of the scope in which it was taken (unless explicitly dropped with `drop` or `Mutex::unlock`).

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 126 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NvgBAgb2Gk65alonu-X:-NvkdWLu6SZU6E4yxKu8:b-q5knxp) ([raw file](https://github.com/starkware-libs/blockifier/blob/4f50c8409dcdbc1ddb860cb3d4f02a0a683a5aa8/crates/blockifier/src/concurrency/scheduler.rs#L126)):*

Done.
</blockquote></details>

Are you sure we want to block the thread until the lock is released?
Are you sure we want to handle the exception here and not use `unwrap()`?

noaov1 (changes requested):
> Reviewed 1 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1795)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @avi-starkware and @barak-b-starkware)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 126 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NvgBAgb2Gk65alonu-X:-Nvl1CUDBG56ASDKm1Hw:b-1ewab2) ([raw file](https://github.com/starkware-libs/blockifier/blob/4f50c8409dcdbc1ddb860cb3d4f02a0a683a5aa8/crates/blockifier/src/concurrency/scheduler.rs#L126)):*

Are you sure we want to block the thread until the lock is released?
Are you sure we want to handle the exception here and not use `unwrap()`?
</blockquote></details>

What do you think we should do in case of a failed attempt to lock?
Try to continue?
Using the `unwrap()` in this case of failed attempt will result in a panic.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 128 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nvk0ZL28OCpOMxakxWY:-Nvl4sxg6sJJMZS0hbB2:b-4xmeux) ([raw file](https://github.com/starkware-libs/blockifier/blob/4f50c8409dcdbc1ddb860cb3d4f02a0a683a5aa8/crates/blockifier/src/concurrency/scheduler.rs#L128)):*

The lock is held until the end of the scope in which it was taken (unless explicitly dropped with `drop` or `Mutex::unlock`).
</blockquote></details>

Do we want to keep it that long? (in case of `next_version_to_validate` for instance)
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 132 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nvl50lV6UiAkFKTRiuR:-Nvl50lV6UiAkFKTRiuS:b45h) ([raw file](https://github.com/starkware-libs/blockifier/blob/c74ead1b9255468623c0bdd3db54bee0863383fe/crates/blockifier/src/concurrency/scheduler.rs#L132)):*
> ```XML
>         if tx_index < self.chunk_size {
>             let mut status =
>                 self.tx_statuses[tx_index].lock().expect("Transaction status is poisoned.");
> ```





_Suggestion:_
```XML
                self.tx_statuses[tx_index].lock().expect("Status of transaction {tx_index} is poisoned.");
```

avi-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1795)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @barak-b-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 126 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NvgBAgb2Gk65alonu-X:-Nvl90SM2vQW3vYogvlA:bqupy9) ([raw file](https://github.com/starkware-libs/blockifier/blob/4f50c8409dcdbc1ddb860cb3d4f02a0a683a5aa8/crates/blockifier/src/concurrency/scheduler.rs#L126)):*

What do you think we should do in case of a failed attempt to lock?
Try to continue?
Using the `unwrap()` in this case of failed attempt will result in a panic.
</blockquote></details>

Okay, you are right. We should block the thread. But I think we might need to switch to RwLock to make it blocked for a minimal amount of time.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 128 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nvk0ZL28OCpOMxakxWY:-Nvl8Mjq7T_cz77QhmPz:bwgxeoh) ([raw file](https://github.com/starkware-libs/blockifier/blob/4f50c8409dcdbc1ddb860cb3d4f02a0a683a5aa8/crates/blockifier/src/concurrency/scheduler.rs#L128)):*

Do we want to keep it that long? (in case of `next_version_to_validate` for instance)
</blockquote></details>

We don't keep it for long. Only 2-3 lines of code...
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 132 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nvl50lV6UiAkFKTRiuR:-NvlHxsG-6G9IyFjo8wg:bpd15qi) ([raw file](https://github.com/starkware-libs/blockifier/blob/c74ead1b9255468623c0bdd3db54bee0863383fe/crates/blockifier/src/concurrency/scheduler.rs#L132)):*
> ```XML
>         if tx_index < self.chunk_size {
>             let mut status =
>                 self.tx_statuses[tx_index].lock().expect("Transaction status is poisoned.");
> ```

This doesn't work. The string "{tx_index}" appears verbatim in the exception.
I did something else (suggested by chatgpt...). Take a look.

barak-b-starkware (commented):
> Reviewed 1 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1795)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @avi-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 132 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nvl50lV6UiAkFKTRiuR:-Nvlbh27AYMsG8BjkoKl:b-vhmfiy) ([raw file](https://github.com/starkware-libs/blockifier/blob/c74ead1b9255468623c0bdd3db54bee0863383fe/crates/blockifier/src/concurrency/scheduler.rs#L132)):*

This doesn't work. The string "{tx_index}" appears verbatim in the exception.
I did something else (suggested by chatgpt...). Take a look.
</blockquote></details>

I think it's time for a `ConcurrencyResult`  or something similar (which will contain `PoisonError` for now).

The return type should be `ConcurrencyResult`, and we should write:
`let mut status = self.tx_statuses[tx_index].lock()?;`
The `lock()` method might return a `PoisonError` which we can propagate (instead of "repeating" it by the `"Status of transaction {tx_index} is poisoned."` string).

BTW, we can match it, and in case of `PoisonError` use `try_lock()` to extract the inner value.

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 132 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nvl50lV6UiAkFKTRiuR:-NwE0d7U0BHZtl50nHpy:b-2eg9pp) ([raw file](https://github.com/starkware-libs/blockifier/blob/c74ead1b9255468623c0bdd3db54bee0863383fe/crates/blockifier/src/concurrency/scheduler.rs#L132)):*

I think it's time for a `ConcurrencyResult`  or something similar (which will contain `PoisonError` for now).

The return type should be `ConcurrencyResult`, and we should write:
`let mut status = self.tx_statuses[tx_index].lock()?;`
The `lock()` method might return a `PoisonError` which we can propagate (instead of "repeating" it by the `"Status of transaction {tx_index} is poisoned."` string).

BTW, we can match it, and in case of `PoisonError` use `try_lock()` to extract the inner value.
</blockquote></details>

After discussing this offline with @noaov1, we concluded that we don't want to define an error type for the scheduler just yet. We want the scheduler to panic in cases where there is unexpected behavior to uncover bugs.

We will need to define `ConcurrencyResult` or `ConcurrencyError` only in cases where we want to recover from said error. In the current case, a `PoisonError` would indicate a bug in the scheduler rather than an error we want to handle.

Instead of `ConcurrencyResult`, I created a method called `lock_tx_status` to avoid code duplication.

@noaov1, please confirm.

barak-b-starkware (changes requested):
> Reviewed 1 of 1 files at r5, 1 of 1 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1795)* status: all files reviewed, 6 unresolved discussions (waiting on @avi-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 85 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwF2pgC6a_8GjdeZ83c:-NwF2pgC6a_8GjdeZ83d:b3eyub0) ([raw file](https://github.com/starkware-libs/blockifier/blob/1a40a0d1c18d5227d2a35a80822cf69f7b08e119/crates/blockifier/src/concurrency/scheduler_test.rs#L85)):*
> ```XML
>     let status = arc_scheduler.lock_tx_status(tx_index);
>     assert_eq!(*status, tx_status);
> }
> ```

I think two separate tests would be better here



_Code quote:_
```XML
#[rstest]
#[case::happy_flow(0, TransactionStatus::ReadyToExecute, false)]
#[should_panic(expected = "Status of transaction index 0 is poisoned: PoisonError { .. }")]
#[case::poisoned(0, TransactionStatus::ReadyToExecute, true)]
fn test_lock_tx_status(
    #[case] tx_index: TxIndex,
    #[case] tx_status: TransactionStatus,
    #[case] poison_thread: bool,
) {
    let arc_scheduler = Arc::new(Scheduler::new(DEFAULT_CHUNK_SIZE));
    if poison_thread {
        let arc_scheduler_clone = arc_scheduler.clone();
        let handle = std::thread::spawn(move || {
            let _guard = arc_scheduler_clone.lock_tx_status(tx_index);
            panic!("Intentional panic to poison the mutex")
        });
        let result = handle.join();
        assert!(result.is_err());
    }
    let status = arc_scheduler.lock_tx_status(tx_index);
    assert_eq!(*status, tx_status);
}
```
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 113 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwFA4PN8p2BfQkVFhU7:-NwFA4PN8p2BfQkVFhU8:b-k3sfg8) ([raw file](https://github.com/starkware-libs/blockifier/blob/1a40a0d1c18d5227d2a35a80822cf69f7b08e119/crates/blockifier/src/concurrency/scheduler_test.rs#L113)):*
> ```XML
>         let mut status = scheduler.tx_statuses[validation_index].lock().unwrap();
>         *status = TransactionStatus::Executed;
>     }
> ```

Can you try adjusting the `default_scheduler!` macro so the `validation_index` & the `execution_index` determine the `tx_statuses` obvious values?
e.g.:
`default_scheduler!(validation_index: 3, execution_index: 4)` creates `tx_statuses` as follows:
`[EXECUTED], [EXECUTED], [EXECUTED], [EXECUTING], [READY_TO_EXECUTE], [READY_TO_EXECUTE], ...`
or something like that...
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 171 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwFBths6oS4SDqdDV3r:-NwFBths6oS4SDqdDV3s:b-pftvn7) ([raw file](https://github.com/starkware-libs/blockifier/blob/1a40a0d1c18d5227d2a35a80822cf69f7b08e119/crates/blockifier/src/concurrency/scheduler_test.rs#L171)):*
> ```XML
>         let mut status = scheduler.tx_statuses[tx_index].lock().unwrap();
>         *status = tx_status;
>     }
> ```

See my comment above re the intialization of the statuses by the `default_scheduler!` macro.



_Code quote:_
```XML
    if let Some(tx_status) = tx_status {
        let mut status = scheduler.tx_statuses[tx_index].lock().unwrap();
        *status = tx_status;
    }
```
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 194 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwFC3236WaEdyHFVgc0:-NwFC3236WaEdyHFVgc1:b-pftvn7) ([raw file](https://github.com/starkware-libs/blockifier/blob/1a40a0d1c18d5227d2a35a80822cf69f7b08e119/crates/blockifier/src/concurrency/scheduler_test.rs#L194)):*
> ```XML
>         let mut status = scheduler.tx_statuses[validation_index].lock().unwrap();
>         *status = tx_status;
>     }
> ```

See my comment above re the intialization of the statuses by the `default_scheduler!` macro.



_Code quote:_
```XML
    if let Some(tx_status) = tx_status {
        let mut status = scheduler.tx_statuses[validation_index].lock().unwrap();
        *status = tx_status;
    }
```
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 218 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwFC4MG-4fV33Aicq7x:-NwFC4MG-4fV33Aicq7y:b-pftvn7) ([raw file](https://github.com/starkware-libs/blockifier/blob/1a40a0d1c18d5227d2a35a80822cf69f7b08e119/crates/blockifier/src/concurrency/scheduler_test.rs#L218)):*
> ```XML
>         let mut status = scheduler.tx_statuses[execution_index].lock().unwrap();
>         *status = tx_status;
>     }
> ```

See my comment above re the intialization of the statuses by the `default_scheduler!` macro.



_Code quote:_
```XML
    if let Some(tx_status) = tx_status {
        let mut status = scheduler.tx_statuses[execution_index].lock().unwrap();
        *status = tx_status;
    }
```

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 85 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwF2pgC6a_8GjdeZ83c:-NwJGy68EK4m8qpHxigI:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/1a40a0d1c18d5227d2a35a80822cf69f7b08e119/crates/blockifier/src/concurrency/scheduler_test.rs#L85)):*

I think two separate tests would be better here
</blockquote></details>

Done.

avi-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1795)* status: 1 of 2 files reviewed, 6 unresolved discussions (waiting on @barak-b-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 113 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwFA4PN8p2BfQkVFhU7:-NwJIUHu6FZC1Kr2fEbQ:b-2z3hgj) ([raw file](https://github.com/starkware-libs/blockifier/blob/1a40a0d1c18d5227d2a35a80822cf69f7b08e119/crates/blockifier/src/concurrency/scheduler_test.rs#L113)):*

Can you try adjusting the `default_scheduler!` macro so the `validation_index` & the `execution_index` determine the `tx_statuses` obvious values?
e.g.:
`default_scheduler!(validation_index: 3, execution_index: 4)` creates `tx_statuses` as follows:
`[EXECUTED], [EXECUTED], [EXECUTED], [EXECUTING], [READY_TO_EXECUTE], [READY_TO_EXECUTE], ...`
or something like that...
</blockquote></details>

I don't think this is a good idea. It will make implicit changes to the setup of each test that would be hard to understand by just looking at the code. All the other field changes possible with this macro are explicit: you input the name of the field and its value directly. However, the changes to the statuses you suggested would not be noticeable to someone reading the code for the first time, and would make the flow of the test harder to understand.

Moreover, I think this is the only test where changing the macro might be useful, and yet I still want to test other flows where the status is not `Executed` for the validation_index, so even here we would have to manually change the status to reach these flows.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 171 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwFBths6oS4SDqdDV3r:-NwJHcTzB16vJLMbdgI7:bd354k7) ([raw file](https://github.com/starkware-libs/blockifier/blob/1a40a0d1c18d5227d2a35a80822cf69f7b08e119/crates/blockifier/src/concurrency/scheduler_test.rs#L171)):*

See my comment above re the intialization of the statuses by the `default_scheduler!` macro.
</blockquote></details>

It doesn't help with this test. The point of this is to check all possible statuses and make sure only one status is the correct status. Regardless of what the default status is, all other statuses would have to be checked, so the number of test cases would remain the same. 
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 194 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwFC3236WaEdyHFVgc0:-NwJIQu6D2Y2i6b2LqGp:bd354k7) ([raw file](https://github.com/starkware-libs/blockifier/blob/1a40a0d1c18d5227d2a35a80822cf69f7b08e119/crates/blockifier/src/concurrency/scheduler_test.rs#L194)):*

See my comment above re the intialization of the statuses by the `default_scheduler!` macro.
</blockquote></details>

It doesn't help with this test. The point of this is to check all possible statuses and make sure only one status is the correct status. Regardless of what the default status is, all other statuses would have to be checked, so the number of test cases would remain the same. 
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 218 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwFC4MG-4fV33Aicq7x:-NwJIRQ70oL1L44Ltp0m:bd354k7) ([raw file](https://github.com/starkware-libs/blockifier/blob/1a40a0d1c18d5227d2a35a80822cf69f7b08e119/crates/blockifier/src/concurrency/scheduler_test.rs#L218)):*

See my comment above re the intialization of the statuses by the `default_scheduler!` macro.
</blockquote></details>

It doesn't help with this test. The point of this is to check all possible statuses and make sure only one status is the correct status. Regardless of what the default status is, all other statuses would have to be checked, so the number of test cases would remain the same.

barak-b-starkware (changes requested):
> Reviewed 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1795)* status: all files reviewed, 4 unresolved discussions (waiting on @avi-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 113 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwFA4PN8p2BfQkVFhU7:-NwTeldRBCmoeW0hDXWq:b-nebvns) ([raw file](https://github.com/starkware-libs/blockifier/blob/1a40a0d1c18d5227d2a35a80822cf69f7b08e119/crates/blockifier/src/concurrency/scheduler_test.rs#L113)):*

I don't think this is a good idea. It will make implicit changes to the setup of each test that would be hard to understand by just looking at the code. All the other field changes possible with this macro are explicit: you input the name of the field and its value directly. However, the changes to the statuses you suggested would not be noticeable to someone reading the code for the first time, and would make the flow of the test harder to understand.

Moreover, I think this is the only test where changing the macro might be useful, and yet I still want to test other flows where the status is not `Executed` for the validation_index, so even here we would have to manually change the status to reach these flows.
</blockquote></details>

OK, I agree. However, I think we should find another way to do that, maybe allowing the macro to get `tx_statuses` or something like that.
We can think about it together when we meet.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 65 at r8](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwTg6M742kfc4M2aYPF:-NwTg6M742kfc4M2aYPG:undefined) ([raw file](https://github.com/starkware-libs/blockifier/blob/68d4c31d5645821867e2263a5aea9fda8baabd90/crates/blockifier/src/concurrency/scheduler_test.rs#L65)):*
> ```XML
> 
> #[rstest]
> fn test_lock_tx_status_no_panic() {
> ```





_Suggestion:_
```XML
fn test_lock_tx_status() {
```
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 84 at r8](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwTh0OqDJEHbzaz9tD5:-NwTh0OqDJEHbzaz9tD6:undefined) ([raw file](https://github.com/starkware-libs/blockifier/blob/68d4c31d5645821867e2263a5aea9fda8baabd90/crates/blockifier/src/concurrency/scheduler_test.rs#L84)):*
> ```XML
>     // The panic is expected here.
>     let status = arc_scheduler.lock_tx_status(0);
>     assert_eq!(*status, TransactionStatus::ReadyToExecute);
> ```





_Suggestion:_
```XML
    let _guard = arc_scheduler.lock_tx_status(0);
```

noaov1 (changes requested):
> Reviewed 1 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1795)* status: all files reviewed, 10 unresolved discussions (waiting on @avi-starkware)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 126 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NvgBAgb2Gk65alonu-X:-NwUsUrRAYKtJ40pqKBn:b-40xfke) ([raw file](https://github.com/starkware-libs/blockifier/blob/4f50c8409dcdbc1ddb860cb3d4f02a0a683a5aa8/crates/blockifier/src/concurrency/scheduler.rs#L126)):*

Okay, you are right. We should block the thread. But I think we might need to switch to RwLock to make it blocked for a minimal amount of time.
</blockquote></details>

Good idea, let's do that. I'm wondering if that can lead to writes starvation. 
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 128 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nvk0ZL28OCpOMxakxWY:-NwUuX8d3exgAXNErH0a:b8bq46i) ([raw file](https://github.com/starkware-libs/blockifier/blob/4f50c8409dcdbc1ddb860cb3d4f02a0a683a5aa8/crates/blockifier/src/concurrency/scheduler.rs#L128)):*

We don't keep it for long. Only 2-3 lines of code...
</blockquote></details>

I see. Great.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 132 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nvl50lV6UiAkFKTRiuR:-NwUuFjK7Tdu6wHNodk8:b7u03hh) ([raw file](https://github.com/starkware-libs/blockifier/blob/c74ead1b9255468623c0bdd3db54bee0863383fe/crates/blockifier/src/concurrency/scheduler.rs#L132)):*

After discussing this offline with @noaov1, we concluded that we don't want to define an error type for the scheduler just yet. We want the scheduler to panic in cases where there is unexpected behavior to uncover bugs.

We will need to define `ConcurrencyResult` or `ConcurrencyError` only in cases where we want to recover from said error. In the current case, a `PoisonError` would indicate a bug in the scheduler rather than an error we want to handle.

Instead of `ConcurrencyResult`, I created a method called `lock_tx_status` to avoid code duplication.

@noaov1, please confirm.
</blockquote></details>

Correct
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 76 at r8](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwUuAFa0EtWxcwuRAbH:-NwUuAFa0EtWxcwuRAbI:bmhnfr) ([raw file](https://github.com/starkware-libs/blockifier/blob/68d4c31d5645821867e2263a5aea9fda8baabd90/crates/blockifier/src/concurrency/scheduler.rs#L76)):*
> ```XML
>                 "Status of transaction index {} is poisoned. Data: {:?}.",
>                 tx_index,
>                 *error.get_ref()
> ```

What information does it consist? Is it interesting?



_Code quote:_
```XML
                *error.get_ref()
```
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 75 at r8](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwV0BNf-odkM-3lGL_F:-NwV0BNf-odkM-3lGL_G:b9mv27x) ([raw file](https://github.com/starkware-libs/blockifier/blob/68d4c31d5645821867e2263a5aea9fda8baabd90/crates/blockifier/src/concurrency/scheduler_test.rs#L75)):*
> ```XML
> fn test_lock_tx_status_poisoned() {
>     let arc_scheduler = Arc::new(Scheduler::new(DEFAULT_CHUNK_SIZE));
>     let arc_scheduler_clone = arc_scheduler.clone();
> ```

Why do we need an `Arc`? Why not use a reference?



_Suggestion:_
```XML
    let scheduler = Arc::new(Scheduler::new(DEFAULT_CHUNK_SIZE));
    let scheduler_copy = arc_scheduler.clone();
```
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 81 at r8](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwV0uv-Anv8gB5eIIIX:-NwV0uv-Anv8gB5eIIIY:b-lj9sx1) ([raw file](https://github.com/starkware-libs/blockifier/blob/68d4c31d5645821867e2263a5aea9fda8baabd90/crates/blockifier/src/concurrency/scheduler_test.rs#L81)):*
> ```XML
>     });
>     let result = handle.join();
>     assert!(result.is_err());
> ```

Can you please verify that we got the expected error? (the relevant panic message)



_Code quote:_
```XML
    assert!(result.is_err());
```
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 89 at r8](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwVJKp8CI2VorQYqdZj:-NwVJKp8CI2VorQYqdZk:b3igmjz) ([raw file](https://github.com/starkware-libs/blockifier/blob/68d4c31d5645821867e2263a5aea9fda8baabd90/crates/blockifier/src/concurrency/scheduler_test.rs#L89)):*
> ```XML
> #[rstest]
> #[case::done(DEFAULT_CHUNK_SIZE, DEFAULT_CHUNK_SIZE, true, true, Task::Done, 0)]
> #[case::no_task(DEFAULT_CHUNK_SIZE, DEFAULT_CHUNK_SIZE, false, true, Task::NoTask, 0)]
> ```

Why does `allow_status_change = true`? Does it matter?



_Code quote:_
```XML
#[case::done(DEFAULT_CHUNK_SIZE, DEFAULT_CHUNK_SIZE, true, true, Task::Done, 0)]
#[case::no_task(DEFAULT_CHUNK_SIZE, DEFAULT_CHUNK_SIZE, false, true, Task::NoTask, 0)]
```
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 90 at r8](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwVIvsyE3lz0WnB2JcG:-NwVIvsyE3lz0WnB2JcH:b78seod) ([raw file](https://github.com/starkware-libs/blockifier/blob/68d4c31d5645821867e2263a5aea9fda8baabd90/crates/blockifier/src/concurrency/scheduler_test.rs#L90)):*
> ```XML
> #[case::done(DEFAULT_CHUNK_SIZE, DEFAULT_CHUNK_SIZE, true, true, Task::Done, 0)]
> #[case::no_task(DEFAULT_CHUNK_SIZE, DEFAULT_CHUNK_SIZE, false, true, Task::NoTask, 0)]
> #[case::no_task_second_flow(DEFAULT_CHUNK_SIZE, 0, false, false, Task::NoTask, 0)]
> ```

Which scenario does it represent?



_Code quote:_
```XML
#[case::no_task_second_flow(DEFAULT_CHUNK_SIZE, 0, false, false, Task::NoTask, 0)]
```
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 114 at r8](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwVIj1G9Imwb0lzGwv3:-NwVIj1G9Imwb0lzGwv4:b-vt1ar5) ([raw file](https://github.com/starkware-libs/blockifier/blob/68d4c31d5645821867e2263a5aea9fda8baabd90/crates/blockifier/src/concurrency/scheduler_test.rs#L114)):*
> ```XML
>         let mut status = scheduler.lock_tx_status(validation_index);
>         *status = TransactionStatus::Executed;
>     }
> ```

Please add a documentation explaining those code lines.



_Code quote:_
```XML
    if allow_status_change
        && validation_index < execution_index
        && validation_index < DEFAULT_CHUNK_SIZE
    {
        let mut status = scheduler.lock_tx_status(validation_index);
        *status = TransactionStatus::Executed;
    }
```

avi-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1795)* status: 0 of 2 files reviewed, 10 unresolved discussions (waiting on @barak-b-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 126 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NvgBAgb2Gk65alonu-X:-NwjRXUc6_MyNRlD5BPt:b-fl9v95) ([raw file](https://github.com/starkware-libs/blockifier/blob/4f50c8409dcdbc1ddb860cb3d4f02a0a683a5aa8/crates/blockifier/src/concurrency/scheduler.rs#L126)):*

Good idea, let's do that. I'm wondering if that can lead to writes starvation.
</blockquote></details>

I added a TODO to consider using RwLock instead of Mutex.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 76 at r8](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwUuAFa0EtWxcwuRAbH:-NwjQioXAKkv7-qnuHCw:bd7gygi) ([raw file](https://github.com/starkware-libs/blockifier/blob/68d4c31d5645821867e2263a5aea9fda8baabd90/crates/blockifier/src/concurrency/scheduler.rs#L76)):*

What information does it consist? Is it interesting?
</blockquote></details>

The error simply contains the data that was in the mutex when it was poisoned.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 113 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwFA4PN8p2BfQkVFhU7:-Nwj_4kt2-ihp4Qmgfar:bgdqi2x) ([raw file](https://github.com/starkware-libs/blockifier/blob/1a40a0d1c18d5227d2a35a80822cf69f7b08e119/crates/blockifier/src/concurrency/scheduler_test.rs#L113)):*

OK, I agree. However, I think we should find another way to do that, maybe allowing the macro to get `tx_statuses` or something like that.
We can think about it together when we meet.
</blockquote></details>

Okay. Let's talk about it when we meet.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 65 at r8](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwTg6M742kfc4M2aYPF:-NwjRukfEpjlyCFwuM96:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/68d4c31d5645821867e2263a5aea9fda8baabd90/crates/blockifier/src/concurrency/scheduler_test.rs#L65)):*
> ```XML
> 
> #[rstest]
> fn test_lock_tx_status_no_panic() {
> ```

Done.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 84 at r8](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwTh0OqDJEHbzaz9tD5:-NwjZd5g7OjEk4jI0ZgG:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/68d4c31d5645821867e2263a5aea9fda8baabd90/crates/blockifier/src/concurrency/scheduler_test.rs#L84)):*
> ```XML
>     // The panic is expected here.
>     let status = arc_scheduler.lock_tx_status(0);
>     assert_eq!(*status, TransactionStatus::ReadyToExecute);
> ```

Done.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 89 at r8](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwVJKp8CI2VorQYqdZj:-Nwj_DYdCJUAFdGurdAO:b2oc8xd) ([raw file](https://github.com/starkware-libs/blockifier/blob/68d4c31d5645821867e2263a5aea9fda8baabd90/crates/blockifier/src/concurrency/scheduler_test.rs#L89)):*

Why does `allow_status_change = true`? Does it matter?
</blockquote></details>

It doesn't matter. I just thought of `true` as the default value.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 90 at r8](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwVIvsyE3lz0WnB2JcG:-NwjZlMICPV5t4MxiuEr:b920xbw) ([raw file](https://github.com/starkware-libs/blockifier/blob/68d4c31d5645821867e2263a5aea9fda8baabd90/crates/blockifier/src/concurrency/scheduler_test.rs#L90)):*

Which scenario does it represent?
</blockquote></details>

The scenario where both next_validation_index and next_execution_index return `None`.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 114 at r8](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwVIj1G9Imwb0lzGwv3:-Nwjb18l7SYKC8gVg66a:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/68d4c31d5645821867e2263a5aea9fda8baabd90/crates/blockifier/src/concurrency/scheduler_test.rs#L114)):*

Please add a documentation explaining those code lines.
</blockquote></details>

Done.

avi-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1795)* status: 0 of 2 files reviewed, 10 unresolved discussions (waiting on @barak-b-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 75 at r8](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwV0BNf-odkM-3lGL_F:-NwnB1jQD2syDj0ovx85:b-czjvlg) ([raw file](https://github.com/starkware-libs/blockifier/blob/68d4c31d5645821867e2263a5aea9fda8baabd90/crates/blockifier/src/concurrency/scheduler_test.rs#L75)):*

Why do we need an `Arc`? Why not use a reference?
</blockquote></details>

As we discussed offline `Arc` is the simplest solution here. It is needed to make sure that the scheduler does not get dropped in the main thread while the spawned thread is still running. Two alternatives:
1. [Use `rayon::scope`:](https://docs.rs/rayon/latest/rayon/fn.scope.html#accessing-stack-data) ensures that all variables outside the scope outlive the scope, and thus a thread can borrow these variables safely. This is the solution used by Aptos in their executor, but it is overkill for this unit test.
2. Use lifetimes to make sure that the scheduler outlives the thread. Lifetimes dirty the code and we prefer to avoid using them.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 81 at r8](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwV0uv-Anv8gB5eIIIX:-NwnB0_U0EDPlLRnvTPX:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/68d4c31d5645821867e2263a5aea9fda8baabd90/crates/blockifier/src/concurrency/scheduler_test.rs#L81)):*

Can you please verify that we got the expected error? (the relevant panic message)
</blockquote></details>

Done.

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 81 at r8](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwV0uv-Anv8gB5eIIIX:-NwnGEDv-DI0CwdD4mmg:b-c0noin) ([raw file](https://github.com/starkware-libs/blockifier/blob/68d4c31d5645821867e2263a5aea9fda8baabd90/crates/blockifier/src/concurrency/scheduler_test.rs#L81)):*

Done.
</blockquote></details>

Actually, the result does not contain the message of the panic...
Anyway this is not important, since the panic is only used to poison the mutex and it is not a functionality we want to test. I did something slightly different. Take a look.

noaov1 (changes requested):
> Reviewed 1 of 1 files at r10.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1795)* status: all files reviewed, 10 unresolved discussions (waiting on @avi-starkware)

___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 160 at r10](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwtKd-r-znKyXoEotm3:-NwtKd-r-znKyXoEotm4:btzo28y) ([raw file](https://github.com/starkware-libs/blockifier/blob/18b3a44012765111cdd9eb7cac3adc3b38adf569/crates/blockifier/src/concurrency/scheduler_test.rs#L160)):*
> ```XML
> #[rstest]
> #[case::from_ready_to_execute_to_executing(0, Some(TransactionStatus::ReadyToExecute), Some(0), 1)]
> #[case::executing(1, Some(TransactionStatus::Executing), None, 0)]
> ```

Why change the `tx_index` from 0 to 1?



_Code quote:_
```XML
#[case::from_ready_to_execute_to_executing(0, Some(TransactionStatus::ReadyToExecute), Some(0), 1)]
#[case::executing(1, Some(TransactionStatus::Executing), None, 0)]
```
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 166 at r10](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwtKVwhC2nP5wksJiay:-NwtKVwi0YSJzp1tECe6:b-z9ggvg) ([raw file](https://github.com/starkware-libs/blockifier/blob/18b3a44012765111cdd9eb7cac3adc3b38adf569/crates/blockifier/src/concurrency/scheduler_test.rs#L166)):*
> ```XML
> fn test_try_incarnate(
>     #[case] tx_index: usize,
>     #[case] tx_status: Option<TransactionStatus>,
> ```

Can we make sure we tested all the possible statuses? 
Consider making it not optional (and set it to `ReadyToExecute` when it does not matter). 



_Code quote:_
```XML
    #[case] tx_status: Option<TransactionStatus>,
```
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 168 at r10](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwtK5gM3khEv2URDkPL:-NwtK5gM3khEv2URDkPM:b-ai87m6) ([raw file](https://github.com/starkware-libs/blockifier/blob/18b3a44012765111cdd9eb7cac3adc3b38adf569/crates/blockifier/src/concurrency/scheduler_test.rs#L168)):*
> ```XML
>     #[case] tx_status: Option<TransactionStatus>,
>     #[case] expected_output: Option<usize>,
>     #[case] expected_n_active_tasks: usize,
> ```

These two are connected, right? Can you avoid one of them?



_Code quote:_
```XML
    #[case] expected_output: Option<usize>,
    #[case] expected_n_active_tasks: usize,
```
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 190 at r10](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwtMTyC2Hp0bWCgU4KS:-NwtMTyC2Hp0bWCgU4KT:bu62c41) ([raw file](https://github.com/starkware-libs/blockifier/blob/18b3a44012765111cdd9eb7cac3adc3b38adf569/crates/blockifier/src/concurrency/scheduler_test.rs#L190)):*
> ```XML
>     #[case] expected_output: Option<usize>,
>     #[case] expected_validation_index: usize,
>     #[case] expected_n_active_tasks: usize,
> ```

Can you use one input and deduce the others?
`expected_output = None` => `expected_n_active_tasks = 0`, no?



_Code quote:_
```XML
    #[case] expected_output: Option<usize>,
    #[case] expected_validation_index: usize,
    #[case] expected_n_active_tasks: usize,
```

avi-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1795)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @barak-b-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 75 at r8](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwV0BNf-odkM-3lGL_F:-NwtvJYp8dASrPFNWXgZ:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/68d4c31d5645821867e2263a5aea9fda8baabd90/crates/blockifier/src/concurrency/scheduler_test.rs#L75)):*

Great. I think you can remove the `arc_` prefix though.
</blockquote></details>

Done.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 90 at r8](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwVIvsyE3lz0WnB2JcG:-NwuI97j2PoVjLtwPeSD:b5fwrhn) ([raw file](https://github.com/starkware-libs/blockifier/blob/68d4c31d5645821867e2263a5aea9fda8baabd90/crates/blockifier/src/concurrency/scheduler_test.rs#L90)):*

Got it. Consider renaming it to `no_task_validation_waiting_for_execution_to_finish` (or a better name you have in mind). Maybe set `validation_index = DEFAULT_CHUNK_SIZE - 1`?
</blockquote></details>

I prefer not to change it to DEFAULT_CHUNK_SIZE-1 because it makes it harder to see that the validation index is smaller.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 80 at r10](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwspfQ_-7mCWB-yAhzo:-NwtvS9BAl65nqNJuYjf:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/18b3a44012765111cdd9eb7cac3adc3b38adf569/crates/blockifier/src/concurrency/scheduler_test.rs#L80)):*
> ```XML
>         panic!("Intentional panic to poison the mutex")
>     });
>     let _result = handle.join().expect_err("Thread did not panic as expected");
> ```

Done.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 90 at r10](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwssmFXB82KVLI8HQRl:-Nwtvo-AC1vso1-eiuey:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/18b3a44012765111cdd9eb7cac3adc3b38adf569/crates/blockifier/src/concurrency/scheduler_test.rs#L90)):*
> ```XML
> #[case::no_task_second_flow(DEFAULT_CHUNK_SIZE, 0, false, false, Task::NoTask, 0)]
> #[case::execution_task(0, 0, false, true, Task::ExecutionTask(0), 1)]
> #[case::execution_task_second_flow(1, 0, false, false, Task::ExecutionTask(1), 1)]
> ```

Done.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 97 at r10](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nwsq1eD1e4LzseA35iP:-NwuFK-H0V0lUJcJ2t8Z:b-872578) ([raw file](https://github.com/starkware-libs/blockifier/blob/18b3a44012765111cdd9eb7cac3adc3b38adf569/crates/blockifier/src/concurrency/scheduler_test.rs#L97)):*

Or something similar, so it will be more informative.
</blockquote></details>

I used Barak's suggestion.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 116 at r10](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nwt52nPD7peGQHDR2K8:-NwuFOw6EfAShxK2YrCy:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/18b3a44012765111cdd9eb7cac3adc3b38adf569/crates/blockifier/src/concurrency/scheduler_test.rs#L116)):*

Add:
```
    #[cfg(test)]
    fn set_tx_status_at_index(&mut self, index: TxIndex, tx_status: TransactionStatus) {
        let mut state = self.lock_tx_status(index);
        *state = tx_status;
    }
```
to the scheduler and
</blockquote></details>

Done.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 160 at r10](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwtKd-r-znKyXoEotm3:-NwtwnRd968kN0aYLxz2:b-wc1s9k) ([raw file](https://github.com/starkware-libs/blockifier/blob/18b3a44012765111cdd9eb7cac3adc3b38adf569/crates/blockifier/src/concurrency/scheduler_test.rs#L160)):*

Why change the `tx_index` from 0 to 1?
</blockquote></details>

No reason. I can keep all at 0.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 166 at r10](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwtKVwhC2nP5wksJiay:-NwuM0Dw-CXbZQ2fDkTY:b-bfnng5) ([raw file](https://github.com/starkware-libs/blockifier/blob/18b3a44012765111cdd9eb7cac3adc3b38adf569/crates/blockifier/src/concurrency/scheduler_test.rs#L166)):*

Can we make sure we tested all the possible statuses? 
Consider making it not optional (and set it to `ReadyToExecute` when it does not matter).
</blockquote></details>

I tried to make sure by listing them all explicitly... Is there a way to check all values of an enum using some functionality of `rstest`?
And it is annoying to not have it be an `Option` because we need a way to not set the status when the index is out of bounds.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 168 at r10](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwtK5gM3khEv2URDkPL:-NwuE0x2DwcySMwt-omc:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/18b3a44012765111cdd9eb7cac3adc3b38adf569/crates/blockifier/src/concurrency/scheduler_test.rs#L168)):*

These two are connected, right? Can you avoid one of them?
</blockquote></details>

Done.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 190 at r10](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-NwtMTyC2Hp0bWCgU4KS:-NwuE255EGHUh3Dl4uwQ:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/18b3a44012765111cdd9eb7cac3adc3b38adf569/crates/blockifier/src/concurrency/scheduler_test.rs#L190)):*

Can you use one input and deduce the others?
`expected_output = None` => `expected_n_active_tasks = 0`, no?
</blockquote></details>

Done.

noaov1 (approved):
> Reviewed 2 of 2 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1795)* status: all files reviewed, 2 unresolved discussions (waiting on @avi-starkware and @barak-b-starkware)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 182 at r11](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nx7Qq6b4Ho70NNAa_4m:-Nx7Qq6b4Ho70NNAa_4n:b8uyr6t) ([raw file](https://github.com/starkware-libs/blockifier/blob/2dfd4d358f35e24aff88c1513a775460be5ed318/crates/blockifier/src/concurrency/scheduler.rs#L182)):*
> ```XML
> 
>     #[cfg(test)]
>     fn set_tx_status(&self, tx_index: TxIndex, status: Option<TransactionStatus>) {
> ```

I don't like it being an option but I don't have a better solution.



_Code quote:_
```XML
    fn set_tx_status(&self, tx_index: TxIndex, status: Option<TransactionStatus>) {
```

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 182 at r11](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nx7Qq6b4Ho70NNAa_4m:-Nx7cY_Q3gppHSD8nA1n:b-fafu21) ([raw file](https://github.com/starkware-libs/blockifier/blob/2dfd4d358f35e24aff88c1513a775460be5ed318/crates/blockifier/src/concurrency/scheduler.rs#L182)):*

I don't like it being an option but I don't have a better solution.
</blockquote></details>

We can remove the option and add:
```
if tx_index < self.chunk_size:
```
What do you think?

barak-b-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1795)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on @avi-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 182 at r11](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nx7Qq6b4Ho70NNAa_4m:-Nx7ngJ35GIe9VbQf6vC:bcd2u0y) ([raw file](https://github.com/starkware-libs/blockifier/blob/2dfd4d358f35e24aff88c1513a775460be5ed318/crates/blockifier/src/concurrency/scheduler.rs#L182)):*

We can remove the option and add:
```
if tx_index < self.chunk_size:
```
What do you think?
</blockquote></details>

Nice! I like it (w/o the `Option`)
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 98 at r12](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nx7mE0IErgQXA5D2KN2:-Nx7mE0IErgQXA5D2KN3:undefined) ([raw file](https://github.com/starkware-libs/blockifier/blob/681a1437d691bfd9975c1dd536392bada89a4122/crates/blockifier/src/concurrency/scheduler_test.rs#L98)):*
> ```XML
>     Task::NoTask
> )]
> #[case::execution_task(0, 0, false, TransactionStatus::Executed, Task::ExecutionTask(0))]
> ```





_Suggestion:_
```XML
#[case::execution_task(0, 0, false, TransactionStatus::ReadyToExecute, Task::ExecutionTask(0))]
```
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 122 at r12](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nx7nxDQB65ucmxofhDe:-Nx7nxDQB65ucmxofhDf:undefined) ([raw file](https://github.com/starkware-libs/blockifier/blob/681a1437d691bfd9975c1dd536392bada89a4122/crates/blockifier/src/concurrency/scheduler_test.rs#L122)):*
> ```XML
>     if validation_index < execution_index && validation_index < DEFAULT_CHUNK_SIZE {
>         scheduler.set_tx_status(validation_index, Some(validation_index_status))
>     }
> ```





_Suggestion:_
```XML
    if validation_index < DEFAULT_CHUNK_SIZE {
        scheduler.set_tx_status(validation_index, Some(validation_index_status))
    }
```

avi-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1795)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @barak-b-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 182 at r11](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nx7Qq6b4Ho70NNAa_4m:-Nx7pmPU8lNV11yedSaX:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/2dfd4d358f35e24aff88c1513a775460be5ed318/crates/blockifier/src/concurrency/scheduler.rs#L182)):*

Nice! I like it (w/o the `Option`)
</blockquote></details>

Done.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 98 at r12](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nx7mE0IErgQXA5D2KN2:-Nx7pnNO9kJNKZYdttX8:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/681a1437d691bfd9975c1dd536392bada89a4122/crates/blockifier/src/concurrency/scheduler_test.rs#L98)):*
> ```XML
>     Task::NoTask
> )]
> #[case::execution_task(0, 0, false, TransactionStatus::Executed, Task::ExecutionTask(0))]
> ```

Done.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 122 at r12](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nx7nxDQB65ucmxofhDe:-Nx7q3ya2rPgnqVtLLE1:b95xa7l) ([raw file](https://github.com/starkware-libs/blockifier/blob/681a1437d691bfd9975c1dd536392bada89a4122/crates/blockifier/src/concurrency/scheduler_test.rs#L122)):*
> ```XML
>     if validation_index < execution_index && validation_index < DEFAULT_CHUNK_SIZE {
>         scheduler.set_tx_status(validation_index, Some(validation_index_status))
>     }
> ```

Actually, we can get rid of the condition entirely!

avi-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1795)* status: 1 of 2 files reviewed, 8 unresolved discussions (waiting on @barak-b-starkware)

___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 119 at r13](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nx7w9s5Bk6z4A7VFewn:-Nx86MTBDXERjScJ2Z4j:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/7601b986485a553b8ea02fe23812d06cb0edd4d2/crates/blockifier/src/concurrency/scheduler_test.rs#L119)):*
> ```XML
>         validation_index: validation_index,
>         done_marker: done_marker,
>     );
> ```

Done.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 135 at r13](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nx7zWDl--hvaWR_wufB:-Nx85AF6EnAqXySTKBJu:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/7601b986485a553b8ea02fe23812d06cb0edd4d2/crates/blockifier/src/concurrency/scheduler_test.rs#L135)):*
> ```XML
> #[case::target_index_eq_validation_index_eq_zero(0, 0)]
> #[case::target_index_gt_validation_index(1, 0)]
> fn test_decrease_validation_index(#[case] target_index: TxIndex, #[case] validation_index: usize) {
> ```

Done.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 150 at r13](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nx7zoejBo1SwXXe-d2p:-Nx851Mw6DLzUdiuUtqj:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/7601b986485a553b8ea02fe23812d06cb0edd4d2/crates/blockifier/src/concurrency/scheduler_test.rs#L150)):*
> ```XML
> #[case::target_index_eq_execution_index_eq_zero(0, 0)]
> #[case::target_index_gt_execution_index(1, 0)]
> fn test_decrease_execution_index(#[case] target_index: TxIndex, #[case] execution_index: usize) {
> ```

Done.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 161 at r13](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nx8-Zn91boyouNcelF6:-Nx86linBTnQCSgUjB-j:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/7601b986485a553b8ea02fe23812d06cb0edd4d2/crates/blockifier/src/concurrency/scheduler_test.rs#L161)):*
> ```XML
> 
> #[rstest]
> #[case::from_ready_to_execute_to_executing(0, TransactionStatus::ReadyToExecute, Some(0))]
> ```

Done.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 169 at r13](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nx8-JLJEqGya652Jaik:-Nx82Vmi4NxZZ4CCqXR_:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/7601b986485a553b8ea02fe23812d06cb0edd4d2/crates/blockifier/src/concurrency/scheduler_test.rs#L169)):*
> ```XML
>     #[case] tx_index: usize,
>     #[case] tx_status: TransactionStatus,
>     #[case] expected_output: Option<usize>,
> ```

Done.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 174 at r13](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nx8-xzdDj9x_6eSdbax:-Nx84xx-AKj8O8BnBTS4:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/7601b986485a553b8ea02fe23812d06cb0edd4d2/crates/blockifier/src/concurrency/scheduler_test.rs#L174)):*
> ```XML
>     scheduler.set_tx_status(tx_index, tx_status);
>     assert_eq!(scheduler.try_incarnate(tx_index), expected_output);
>     let expected_n_active_tasks = if expected_output.is_some() { 1 } else { 0 };
> ```

Done.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 179 at r13](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nx8-gE3E8SbPX6LIVWU:-Nx82p194yUVDDswncJk:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/7601b986485a553b8ea02fe23812d06cb0edd4d2/crates/blockifier/src/concurrency/scheduler_test.rs#L179)):*

same in `next_version_to_execute`
</blockquote></details>

Done.
___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 187 at r13](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nx8-kk9ATyqx0GmHxMj:-Nx86ru74rKuyNPRCt0d:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/7601b986485a553b8ea02fe23812d06cb0edd4d2/crates/blockifier/src/concurrency/scheduler_test.rs#L187)):*

same in `next_version_to_execute`
</blockquote></details>

Done.

barak-b-starkware (approved):
:lgtm:

> Reviewed 1 of 1 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-:-Nx8AyUw6VJ6hhJUMw1h:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @avi-starkware)

noaov1 (commented):
> Reviewed 1 of 2 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1795)* status: all files reviewed, 1 unresolved discussion (waiting on @avi-starkware)

___
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 102 at r14](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nx8IWtU32r4eO1OugHx:-Nx8IWtV8Dg4uTJHeaxP:blnlts7) ([raw file](https://github.com/starkware-libs/blockifier/blob/6f69c76abc34e485e31488928115f38cbde056cb/crates/blockifier/src/concurrency/scheduler_test.rs#L102)):*
> ```XML
>     #[case] execution_index: TxIndex,
>     #[case] validation_index: TxIndex,
>     #[case] validation_index_status: TransactionStatus,
> ```

Why do we only set the status in the validation index and not in the execution index?
Don't we want to test the case where `execution_index >= validation_index` but the relevant transaction is not ready to be executed?



_Code quote:_
```XML
    #[case] validation_index_status: TransactionStatus,
```

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler_test.rs` line 102 at r14](https://reviewable.io/reviews/starkware-libs/blockifier/1795#-Nx8IWtU32r4eO1OugHx:-Nx8KoH71q-Zdy2CmFA_:b-4ew98q) ([raw file](https://github.com/starkware-libs/blockifier/blob/6f69c76abc34e485e31488928115f38cbde056cb/crates/blockifier/src/concurrency/scheduler_test.rs#L102)):*

Why do we only set the status in the validation index and not in the execution index?
Don't we want to test the case where `execution_index >= validation_index` but the relevant transaction is not ready to be executed?
</blockquote></details>

This case is tested in the unit test for `next_version_to_execute`. Within the method `next_task` the flow is the same whether the status is ReadyToExecute or not.

noaov1 (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1795)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @avi-starkware)

PR #1830, opened 2024-04-27
test(concurrency): multithreaded flow test for the scheduler

barak-b-starkware:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)

### DIFF
--- a/crates/blockifier/src/abi/sierra_types.rs
+++ b/crates/blockifier/src/abi/sierra_types.rs
@@ -36,16 +36,21 @@ pub trait SierraType: Sized {
     fn from_memory(vm: &VirtualMachine, ptr: &mut Relocatable) -> SierraTypeResult<Self>;
 
     fn from_storage(
-        state: &mut dyn StateReader,
+        state: &dyn StateReader,
         contract_address: &ContractAddress,
         key: &StorageKey,
     ) -> SierraTypeResult<Self>;
 }
 
 // Utils.
+
 pub fn felt_to_u128(felt: &Felt252) -> Result<u128, SierraTypeError> {
     felt.to_u128()
         .ok_or_else(|| SierraTypeError::ValueTooLargeForType { val: felt.clone(), ty: "u128" })
+}
+
+pub fn stark_felt_to_u128(stark_felt: &StarkFelt) -> Result<u128, SierraTypeError> {
+    felt_to_u128(&stark_felt_to_felt(*stark_felt))
 }
 
 // TODO(barak, 01/10/2023): Move to starknet_api under StorageKey implementation.
@@ -78,7 +83,7 @@ impl SierraType for SierraU128 {
     }
 
     fn from_storage(
-        state: &mut dyn StateReader,
+        state: &dyn StateReader,
         contract_address: &ContractAddress,
         key: &StorageKey,
     ) -> SierraTypeResult<Self> {
@@ -111,7 +116,7 @@ impl SierraType for SierraU256 {
     }
 
     fn from_storage(
-        state: &mut dyn StateReader,
+        state: &dyn StateReader,
         contract_address: &ContractAddress,
         key: &StorageKey,
     ) -> SierraTypeResult<Self> {
--- a/crates/blockifier/src/concurrency/flow_test.rs
+++ b/crates/blockifier/src/concurrency/flow_test.rs
@@ -1,0 +1,174 @@
+use std::collections::HashMap;
+use std::sync::atomic::Ordering;
+use std::sync::Arc;
+
+use rstest::rstest;
+use starknet_api::core::{ContractAddress, PatriciaKey};
+use starknet_api::hash::{StarkFelt, StarkHash};
+use starknet_api::{contract_address, patricia_key, stark_felt};
+
+use crate::abi::sierra_types::{SierraType, SierraU128};
+use crate::concurrency::scheduler::{Scheduler, Task, TransactionStatus};
+use crate::concurrency::test_utils::{safe_versioned_state_for_testing, DEFAULT_CHUNK_SIZE};
+use crate::concurrency::versioned_state::ThreadSafeVersionedState;
+use crate::state::cached_state::{CachedState, ContractClassMapping, StateMaps};
+use crate::state::state_api::UpdatableState;
+use crate::storage_key;
+use crate::test_utils::dict_state_reader::DictStateReader;
+
+const CONTRACT_ADDRESS: &str = "0x18031991";
+const STORAGE_KEY: u8 = 27;
+
+#[rstest]
+fn scheduler_flow_test(
+    // TODO(barak, 01/07/2024): Add a separate identical test and use the package loom.
+    #[values(1, 2, 4, 32, 64, 128)] num_threads: u8,
+) {
+    // Tests the Scheduler under a heavy load of validation aborts. To do that, we simulate multiple
+    // transactions with multiple threads, where every transaction depends on its predecessor. Each
+    // transaction sequentially advances a counter by reading the previous value and bumping it by
+    // 1.
+    let scheduler = Arc::new(Scheduler::new(DEFAULT_CHUNK_SIZE));
+    let versioned_state =
+        safe_versioned_state_for_testing(CachedState::from(DictStateReader::default()));
+    let mut handles = vec![];
+
+    for _ in 0..num_threads {
+        let scheduler = Arc::clone(&scheduler);
+        let versioned_state = versioned_state.clone();
+        let handle = std::thread::spawn(move || {
+            let mut task = Task::NoTask;
+            loop {
+                if let Some(mut transaction_committer) = scheduler.try_enter_commit_phase() {
+                    while let Some(tx_index) = transaction_committer.try_commit() {
+                        let mut state_proxy = versioned_state.pin_version(tx_index);
+                        let (reads, writes) =
+                            get_reads_writes_for(Task::ValidationTask(tx_index), &versioned_state);
+                        let reads_valid = state_proxy.validate_reads(&reads);
+                        if !reads_valid {
+                            state_proxy.delete_writes(&writes, &ContractClassMapping::default());
+                            let (_, new_writes) = get_reads_writes_for(
+                                Task::ExecutionTask(tx_index),
+                                &versioned_state,
+                            );
+                            state_proxy.apply_writes(
+                                &new_writes,
+                                &ContractClassMapping::default(),
+                                &HashMap::default(),
+                            );
+                            scheduler.finish_execution_during_commit(tx_index);
+                        }
+                    }
+                }
+                task = match task {
+                    Task::ExecutionTask(tx_index) => {
+                        let (_, writes) =
+                            get_reads_writes_for(Task::ExecutionTask(tx_index), &versioned_state);
+                        versioned_state.pin_version(tx_index).apply_writes(
+                            &writes,
+                            &ContractClassMapping::default(),
+                            &HashMap::default(),
+                        );
+                        scheduler.finish_execution(tx_index);
+                        Task::NoTask
+                    }
+                    Task::ValidationTask(tx_index) => {
+                        let state_proxy = versioned_state.pin_version(tx_index);
+                        let (reads, writes) =
+                            get_reads_writes_for(Task::ValidationTask(tx_index), &versioned_state);
+                        let read_set_valid = state_proxy.validate_reads(&reads);
+                        let aborted = !read_set_valid && scheduler.try_validation_abort(tx_index);
+                        if aborted {
+                            state_proxy.delete_writes(&writes, &ContractClassMapping::default());
+                            scheduler.finish_abort(tx_index)
+                        } else {
+                            Task::NoTask
+                        }
+                    }
+                    Task::NoTask => scheduler.next_task(),
+                    Task::Done => break,
+                }
+            }
+        });
+        handles.push(handle);
+    }
+    for handle in handles {
+        handle.join().unwrap();
+    }
+
+    // The execution index can be strictly greater than chunk_size. This is a side effect of using
+    // atomic variables instead of locks, which can encapsulate both the check (whether to increment
+    // the variable or not) and the incrementation in a scope where no other threads can access the
+    // variable.
+    assert!(scheduler.execution_index.load(Ordering::Acquire) >= DEFAULT_CHUNK_SIZE);
+    // There is no guarantee about the validation index because of the use of the commit index.
+    assert!(*scheduler.commit_index.lock().unwrap() == DEFAULT_CHUNK_SIZE);
+    assert!(scheduler.get_n_committed_txs() == DEFAULT_CHUNK_SIZE);
+    assert!(scheduler.done_marker.load(Ordering::Acquire));
+    let inner_versioned_state = versioned_state.into_inner_state();
+    for tx_index in 0..DEFAULT_CHUNK_SIZE {
+        assert_eq!(*scheduler.tx_statuses[tx_index].lock().unwrap(), TransactionStatus::Committed);
+        let storage_writes = inner_versioned_state.get_writes_of_index(tx_index).storage;
+        assert_eq!(
+            *storage_writes
+                .get(&(contract_address!(CONTRACT_ADDRESS), storage_key!(STORAGE_KEY)))
+                .unwrap(),
+            stark_felt!(format!("{:x}", tx_index + 1).as_str())
+        );
+    }
+}
+
+fn get_reads_writes_for(
+    task: Task,
+    versioned_state: &ThreadSafeVersionedState<CachedState<DictStateReader>>,
+) -> (StateMaps, StateMaps) {
+    match task {
+        Task::ExecutionTask(tx_index) => {
+            let state_proxy = match tx_index {
+                0 => {
+                    return (
+                        state_maps_with_single_storage_entry(0),
+                        state_maps_with_single_storage_entry(1),
+                    );
+                }
+                _ => versioned_state.pin_version(tx_index - 1),
+            };
+            let tx_written_value = SierraU128::from_storage(
+                &state_proxy,
+                &contract_address!(CONTRACT_ADDRESS),
+                &storage_key!(STORAGE_KEY),
+            )
+            .unwrap()
+            .as_value();
+            (
+                state_maps_with_single_storage_entry(tx_written_value),
+                state_maps_with_single_storage_entry(tx_written_value + 1),
+            )
+        }
+        Task::ValidationTask(tx_index) => {
+            let state_proxy = versioned_state.pin_version(tx_index);
+            let tx_written_value = SierraU128::from_storage(
+                &state_proxy,
+                &contract_address!(CONTRACT_ADDRESS),
+                &storage_key!(STORAGE_KEY),
+            )
+            .unwrap()
+            .as_value();
+            (
+                state_maps_with_single_storage_entry(tx_written_value - 1),
+                state_maps_with_single_storage_entry(tx_written_value),
+            )
+        }
+        _ => panic!("Only execution and validation tasks shold be used here."),
+    }
+}
+
+fn state_maps_with_single_storage_entry(value: u128) -> StateMaps {
+    StateMaps {
+        storage: HashMap::from([(
+            (contract_address!(CONTRACT_ADDRESS), storage_key!(STORAGE_KEY)),
+            stark_felt!(value),
+        )]),
+        ..Default::default()
+    }
+}
--- a/crates/blockifier/src/concurrency/scheduler.rs
+++ b/crates/blockifier/src/concurrency/scheduler.rs
@@ -1,4 +1,3 @@
-use std::cmp::min;
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::{Mutex, MutexGuard, TryLockError};
 
@@ -8,6 +7,10 @@ use crate::concurrency::TxIndex;
 #[cfg(test)]
 #[path = "scheduler_test.rs"]
 pub mod test;
+
+#[cfg(test)]
+#[path = "flow_test.rs"]
+pub mod flow_test;
 
 pub struct TransactionCommitter<'a> {
     scheduler: &'a Scheduler,
@@ -87,10 +90,6 @@ impl Scheduler {
 
         let index_to_validate = self.validation_index.load(Ordering::Acquire);
         let index_to_execute = self.execution_index.load(Ordering::Acquire);
-
-        if min(index_to_validate, index_to_execute) >= self.chunk_size {
-            return Task::NoTask;
-        }
 
         if index_to_validate < index_to_execute {
             if let Some(tx_index) = self.next_version_to_validate() {
--- a/crates/blockifier/src/concurrency/scheduler_test.rs
+++ b/crates/blockifier/src/concurrency/scheduler_test.rs
@@ -6,10 +6,9 @@ use pretty_assertions::assert_eq;
 use rstest::rstest;
 
 use crate::concurrency::scheduler::{Scheduler, Task, TransactionStatus};
+use crate::concurrency::test_utils::DEFAULT_CHUNK_SIZE;
 use crate::concurrency::TxIndex;
 use crate::default_scheduler;
-
-const DEFAULT_CHUNK_SIZE: usize = 100;
 
 #[rstest]
 fn test_new(#[values(0, 1, 32)] chunk_size: usize) {
--- a/crates/blockifier/src/concurrency/test_utils.rs
+++ b/crates/blockifier/src/concurrency/test_utils.rs
@@ -12,6 +12,10 @@ use crate::test_utils::dict_state_reader::DictStateReader;
 use crate::transaction::account_transaction::AccountTransaction;
 use crate::transaction::transactions::ExecutableTransaction;
 
+// Public Consts.
+
+pub const DEFAULT_CHUNK_SIZE: usize = 64;
+
 // Fixtures.
 
 #[fixture]
@@ -54,12 +58,16 @@ macro_rules! default_scheduler {
     };
 }
 
+// Concurrency constructors.
+
 // TODO(meshi, 01/06/2024): Consider making this a macro.
 pub fn safe_versioned_state_for_testing(
     block_state: CachedState<DictStateReader>,
 ) -> ThreadSafeVersionedState<CachedState<DictStateReader>> {
     ThreadSafeVersionedState::new(VersionedState::new(block_state))
 }
+
+// Utils.
 
 // Note: this function does not mutate the state.
 pub fn create_fee_transfer_call_info<S: StateReader>(


barak-b-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @avi-starkware, @noaov1, and @Yoni-Starkware)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 69 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyKszR64ckIaSvjUtbL:-NyKszR76dsTTZ_lQxen:baqzqau) ([raw file](https://github.com/starkware-libs/blockifier/blob/63c8b7e801aba3556790b548bd9044bf1fa9d7d2/crates/blockifier/src/concurrency/scheduler.rs#L69)):*
> ```XML
>         if min(index_to_validate, index_to_execute) >= self.chunk_size {
>             return Task::NoTask;
>         }
> ```

Bug fix (removing this [optimization](https://reviewable.io/reviews/starkware-libs/blockifier/1700#-NvQTp0Z62jPblDJHXOC)).



_Code quote:_
```XML
        if min(index_to_validate, index_to_execute) >= self.chunk_size {
            return Task::NoTask;
        }
```

Yoni-Starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @avi-starkware, @barak-b-starkware, and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 69 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyKszR64ckIaSvjUtbL:-NyLjgUn2_dnVKRl3JJK:btm02s7) ([raw file](https://github.com/starkware-libs/blockifier/blob/63c8b7e801aba3556790b548bd9044bf1fa9d7d2/crates/blockifier/src/concurrency/scheduler.rs#L69)):*

Bug fix (removing this [optimization](https://reviewable.io/reviews/starkware-libs/blockifier/1700#-NvQTp0Z62jPblDJHXOC)).
</blockquote></details>

Can you explain the bug? 
I agree this optimization  should be removed though (negligible)

avi-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @barak-b-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 31 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyMMyiS8pLE1dU_ak-A:-NyMMyiS8pLE1dU_ak-B:b-1yk8ys) ([raw file](https://github.com/starkware-libs/blockifier/blob/8fdceb738c8ce0edb62c898f5cb1d75ea79d0ae4/crates/blockifier/src/concurrency/flow_test.rs#L31)):*
> ```XML
>     let scheduler = Arc::new(Scheduler::new(DEFAULT_CHUNK_SIZE));
>     let storage_key = storage_key!(27_u8);
>     let versioned_state = safe_versioned_state_for_testing(DictStateReader::default());
> ```

Do we need a versioned_state to test the scheduler?
You can predetermine the values of `read_set_valid` instead of actually checking the storage when you get a validation task, and just run `finish_execution` when you get an execution task.

This might make it easier to control the number of invalidations in each of the different cases of this test.



_Code quote:_
```XML
    let versioned_state = safe_versioned_state_for_testing(DictStateReader::default());
```
___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 89 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyL2wGnAruahE-F3_bM:-NyMJfBX9_rdPlozrCmC:b-cj4bmt) ([raw file](https://github.com/starkware-libs/blockifier/blob/8fdceb738c8ce0edb62c898f5cb1d75ea79d0ae4/crates/blockifier/src/concurrency/flow_test.rs#L89)):*

Notice that the `validation_index` can be strictly greater than `chunk_size` at the end of the chunk's execution.
On one hand, IIUC, it's not a bug because it doesn't cause any unexpected behavior (**at least for now**).
On the other hand, it's not what we would expect. We expect to get `validation_index == chunk_size` at the end of the chunk's execution (as happens with `execution_index`).
</blockquote></details>

This is indeed weird, but it is a side effect of using atomic variables instead of locks.
The only place where we increase the validation_index appears immediately after a check that the `validation_index` is lower than `chunk_size`, so we expect `validatoin_index` to never exceed `chunk_size`, but it is not a fool-proof guarantee since two threads might pass the check at the same time with `validation_index = chunk_size - 1` and then both increment the counter making it `chunk_size + 1`.

This can happen for `execution_index` as well.

I think we should always check `>= chunk_size` and not `==chunk_size`.

Your assert for the `execution_index` might sporadically fail.

barak-b-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @avi-starkware, @noaov1, and @Yoni-Starkware)

___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 31 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyMMyiS8pLE1dU_ak-A:-NyOxEbQ8Ocwnb9Zp0uy:b-hy6ci7) ([raw file](https://github.com/starkware-libs/blockifier/blob/8fdceb738c8ce0edb62c898f5cb1d75ea79d0ae4/crates/blockifier/src/concurrency/flow_test.rs#L31)):*

Do we need a versioned_state to test the scheduler?
You can predetermine the values of `read_set_valid` instead of actually checking the storage when you get a validation task, and just run `finish_execution` when you get an execution task.

This might make it easier to control the number of invalidations in each of the different cases of this test.
</blockquote></details>

It's a good point, and I thought about it when I added the `versioed_state,` but I couldn't figure out any good way to do that.
How would you predetermine the `read_set_valid`?
1. Always `true` -> `try_validation_abort` would not be checked, and we would always get `aborted == false`. Notice that if we write `scheduler.try_validation_abort(tx_index) && !read_set_valid` instead of `!read_set_valid && scheduler.try_validation_abort(tx_index)`, `try_validation_abort` would always set the status `Aborting` which is a bug.
2. Always `false` -> `next_version_to_validate` will always return `0`, and the test will run into a timeout.
3. Random? -> I don't like it because it makes things much harder to check later, especially in multithreaded tests. Moreover, I'm unsure what the expected behavior is in each case.

In summary, adding the `versioned_state` here is a constraint, not a planned part. If you have any good ideas on how to get rid of it, it would be great.
BTW that's why I implemented the `delete_writes` as part of this task in a separate PR.
___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 89 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyL2wGnAruahE-F3_bM:-NyP85dH5WdCE10mCUJw:b9tzpyc) ([raw file](https://github.com/starkware-libs/blockifier/blob/8fdceb738c8ce0edb62c898f5cb1d75ea79d0ae4/crates/blockifier/src/concurrency/flow_test.rs#L89)):*

This is indeed weird, but it is a side effect of using atomic variables instead of locks.
The only place where we increase the validation_index appears immediately after a check that the `validation_index` is lower than `chunk_size`, so we expect `validatoin_index` to never exceed `chunk_size`, but it is not a fool-proof guarantee since two threads might pass the check at the same time with `validation_index = chunk_size - 1` and then both increment the counter making it `chunk_size + 1`.

This can happen for `execution_index` as well.

I think we should always check `>= chunk_size` and not `==chunk_size`.

Your assert for the `execution_index` might sporadically fail.
</blockquote></details>

You're right! Changed and documented.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 69 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyKszR64ckIaSvjUtbL:-NyPBBqy1Jq1jKsJIVXj:b-ib0tqr) ([raw file](https://github.com/starkware-libs/blockifier/blob/63c8b7e801aba3556790b548bd9044bf1fa9d7d2/crates/blockifier/src/concurrency/scheduler.rs#L69)):*

Can you explain the bug? 
I agree this optimization  should be removed though (negligible)
</blockquote></details>

`next_version_to_execute` and `next_version_to_validate` are the ones that call `check_done()`.
Assume that `CHUNK_SIZE = 64`, if `validation_index == execution_index == 64`  when you call `next_task` this removed `if` statement will immediately return `Task:NoTask` where the desired behavior is to call `next_version_to_...` that will be evaluated to `None` after he calls `check_done()` and only then the `next_task()` method will reach its last line and will return `Task:NoTask`.

Yoni-Starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @avi-starkware, @barak-b-starkware, and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 69 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyKszR64ckIaSvjUtbL:-NyPMt312GwM4FDSGo07:b-easfw3) ([raw file](https://github.com/starkware-libs/blockifier/blob/63c8b7e801aba3556790b548bd9044bf1fa9d7d2/crates/blockifier/src/concurrency/scheduler.rs#L69)):*

`next_version_to_execute` and `next_version_to_validate` are the ones that call `check_done()`.
Assume that `CHUNK_SIZE = 64`, if `validation_index == execution_index == 64`  when you call `next_task` this removed `if` statement will immediately return `Task:NoTask` where the desired behavior is to call `next_version_to_...` that will be evaluated to `None` after he calls `check_done()` and only then the `next_task()` method will reach its last line and will return `Task:NoTask`.
</blockquote></details>

Oh, and check_done changes the marker. Thanks

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 69 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyKszR64ckIaSvjUtbL:-NyPDkIQ3coRthwz2GW8:b9t8e4w) ([raw file](https://github.com/starkware-libs/blockifier/blob/63c8b7e801aba3556790b548bd9044bf1fa9d7d2/crates/blockifier/src/concurrency/scheduler.rs#L69)):*

`next_version_to_execute` and `next_version_to_validate` are the ones that call `check_done()`.
Assume that `CHUNK_SIZE = 64`, if `validation_index == execution_index == 64`  when you call `next_task` this removed `if` statement will immediately return `Task:NoTask` where the desired behavior is to call `next_version_to_...` that will be evaluated to `None` after he calls `check_done()` and only then the `next_task()` method will reach its last line and will return `Task:NoTask`.
</blockquote></details>

You are correct, but we are removing `check_done()`. The `done_marker` will be activated when all transactions have been committed.

I am fine with removing this optimization anyway.

avi-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @barak-b-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 31 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyMMyiS8pLE1dU_ak-A:-NyPO2Rx6YjdmaF16JCR:b11pl1f) ([raw file](https://github.com/starkware-libs/blockifier/blob/8fdceb738c8ce0edb62c898f5cb1d75ea79d0ae4/crates/blockifier/src/concurrency/flow_test.rs#L31)):*

It's a good point, and I thought about it when I added the `versioed_state,` but I couldn't figure out any good way to do that.
How would you predetermine the `read_set_valid`?
1. Always `true` -> `try_validation_abort` would not be checked, and we would always get `aborted == false`. Notice that if we write `scheduler.try_validation_abort(tx_index) && !read_set_valid` instead of `!read_set_valid && scheduler.try_validation_abort(tx_index)`, `try_validation_abort` would always set the status `Aborting` which is a bug.
2. Always `false` -> `next_version_to_validate` will always return `0`, and the test will run into a timeout.
3. Random? -> I don't like it because it makes things much harder to check later, especially in multithreaded tests. Moreover, I'm unsure what the expected behavior is in each case.

In summary, adding the `versioned_state` here is a constraint, not a planned part. If you have any good ideas on how to get rid of it, it would be great.
BTW that's why I implemented the `delete_writes` as part of this task in a separate PR.
</blockquote></details>

On second thought, I like this test. It creates maximal dependency and checks the scheduler's performance when it eventually has to run all transactions sequentially.
In the future we will also be able to reduce the level of dependency as much as we like, by having `k` storage keys and making transaction `i` write to storage key `i mod k`.

avi-starkware (commented):
> Reviewed 4 of 5 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)* status: all files reviewed, 3 unresolved discussions (waiting on @barak-b-starkware and @noaov1)

Yoni-Starkware (changes requested):
> Reviewed 4 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)* status: all files reviewed, 5 unresolved discussions (waiting on @barak-b-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 63 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyPdlYJ7cStpTWtaP6N:-NyPdlYJ7cStpTWtaP6O:bpm1m6e) ([raw file](https://github.com/starkware-libs/blockifier/blob/83d6c532725ab1465e0649c3cd1e5b79d1df8909/crates/blockifier/src/concurrency/flow_test.rs#L63)):*
> ```XML
>                             contract_address,
>                             storage_key,
>                         );
> ```

Let's make it a bit more noisy - the threads run advance_counter logic, so there should be much more invalidations



_Suggestion:_
```XML
                task = match task {
                    Task::ExecutionTask(tx_index) => {
                        // Advance counter.
                        let state_proxy = versioned_state.pin_version(tx_index);
                        let current_value = state_proxy
                            .get_storage_at(contract_address, storage_key)
                            .unwrap();                        
                        let writes = state_maps_with_single_storage_entry(
                            contract_address,
                            storage_key,
                            current_value + 1
                        );
                        state_proxy.apply_writes(&writes, &ContractClassMapping::default());
                        scheduler.finish_execution(tx_index);
                        Task::NoTask
                    }
                    Task::ValidationTask(tx_index) => {
                        // Get the value written by this transaction (note the `+ 1`).
                        let tx_write_value = versioned_state.pin_version(tx_index + 1)
                            .get_storage_at(contract_address, storage_key)
                            .unwrap();
                        let expected_prev_value = tx_write_value - 1;
                        let read_set = create_trivial_read_set_from_value(
                            current_cell_value,
                            contract_address,
                            expected_prev_value,
                        );
```
___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 124 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyPa0vS-NNQmT3755-x:-NyPa0vS-NNQmT3755-y:undefined) ([raw file](https://github.com/starkware-libs/blockifier/blob/83d6c532725ab1465e0649c3cd1e5b79d1df8909/crates/blockifier/src/concurrency/flow_test.rs#L124)):*
> ```XML
>     value: StarkFelt,
>     contract_address: ContractAddress,
>     storage_key: StorageKey,
> ```





_Suggestion:_
```XML
fn create_trivial_read_set_from_value(
    contract_address: ContractAddress,
    storage_key: StorageKey,
    value: StarkFelt,
```

barak-b-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)* status: all files reviewed, 3 unresolved discussions (waiting on @avi-starkware, @noaov1, and @Yoni-Starkware)

___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 63 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyPdlYJ7cStpTWtaP6N:-NyUVgy8Et0FhaMzIxHf:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/83d6c532725ab1465e0649c3cd1e5b79d1df8909/crates/blockifier/src/concurrency/flow_test.rs#L63)):*

Let's make it a bit more noisy - the threads run advance_counter logic, so there should be much more invalidations
</blockquote></details>

Done.
___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 90 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyPEPi4Ew6CnaYoXRWc:-NyUUFfIA4mPtZFd4kNu:b6nymoh) ([raw file](https://github.com/starkware-libs/blockifier/blob/83d6c532725ab1465e0649c3cd1e5b79d1df8909/crates/blockifier/src/concurrency/flow_test.rs#L90)):*

The check and the incrementation are two atomic operations. The gap allows two threads to pass the check when `validation_index = chunk_size - 1`.
</blockquote></details>

Done.
AFAIK "atomic operation" is a concept that something happens in "a scope where no other threads can access". And not something that preserved to "check" or "incrementation", isn't it?
___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 124 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyPa0vS-NNQmT3755-x:-NyUVdAA2H9tPLmXzrQT:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/83d6c532725ab1465e0649c3cd1e5b79d1df8909/crates/blockifier/src/concurrency/flow_test.rs#L124)):*
> ```XML
>     value: StarkFelt,
>     contract_address: ContractAddress,
>     storage_key: StorageKey,
> ```

Done.

avi-starkware:
*[`crates/blockifier/src/concurrency/flow_test.rs` line 90 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyPEPi4Ew6CnaYoXRWc:-NyUWsrPE3GlSZ2Vt1rV:b-jivgmy) ([raw file](https://github.com/starkware-libs/blockifier/blob/83d6c532725ab1465e0649c3cd1e5b79d1df8909/crates/blockifier/src/concurrency/flow_test.rs#L90)):*

Done.
AFAIK "atomic operation" is a concept that something happens in "a scope where no other threads can access". And not something that preserved to "check" or "incrementation", isn't it?
</blockquote></details>

I just think that when discussing atomic variables, "atomic operations" evokes "operations on atomic variables", and not "operations in a scope where no other thread can access" so it is a bit confusing.
By the way, I think atomic operations actually mean operations that happen in a single CPU step and thus cannot be split.

avi-starkware:
*[`crates/blockifier/src/concurrency/flow_test.rs` line 77 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyVYhX4AhGcM8--11sx:-NyVYhX4AhGcM8--11sy:undefined) ([raw file](https://github.com/starkware-libs/blockifier/blob/39d59e88c3591941a74f75429b4baf417e5dd44b/crates/blockifier/src/concurrency/flow_test.rs#L77)):*
> ```XML
>                         .unwrap();
>                         // If tx number tx_index wrote tx_written_value then it must be read
>                         // tx_written_value - 1.
> ```





_Suggestion:_
```XML
                        // If tx number tx_index wrote tx_written_value then it must have read
                        // tx_written_value - 1.
```

avi-starkware:
*[`crates/blockifier/src/concurrency/flow_test.rs` line 94 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyVZ-gb0MIEInsdSFlL:-NyVZ-gb0MIEInsdSFlM:b-b4p365) ([raw file](https://github.com/starkware-libs/blockifier/blob/39d59e88c3591941a74f75429b4baf417e5dd44b/crates/blockifier/src/concurrency/flow_test.rs#L94)):*
> ```XML
>                             state_proxy.delete_writes(&write_set, &ContractClassMapping::default());
>                         }
>                         scheduler.finish_validation(tx_index, aborted)
> ```

We changed the method `finish_validation` to `finish_abort` (to be used only if `aborted == true`)



_Suggestion:_
```XML
                            scheduler.finish_abort(tx_index)
                        } else {
                            Task::NoTask
                        }
```

barak-b-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)* status: all files reviewed, 8 unresolved discussions (waiting on @avi-starkware, @noaov1, and @Yoni-Starkware)

___
*[`crates/blockifier/src/abi/sierra_types.rs` line 39 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyUyWke46wwBCmm9jl1:-NyWEVMPAp_flnovfPFb:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/39d59e88c3591941a74f75429b4baf417e5dd44b/crates/blockifier/src/abi/sierra_types.rs#L39)):*
> ```XML
> 
>     fn from_storage(
>         state: &mut dyn StateReader,
> ```

Done.
___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 90 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyPEPi4Ew6CnaYoXRWc:-NyWGphj552RUZbzjiRe:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/83d6c532725ab1465e0649c3cd1e5b79d1df8909/crates/blockifier/src/concurrency/flow_test.rs#L90)):*

I just think that when discussing atomic variables, "atomic operations" evokes "operations on atomic variables", and not "operations in a scope where no other thread can access" so it is a bit confusing.
By the way, I think atomic operations actually mean operations that happen in a single CPU step and thus cannot be split.
</blockquote></details>

Done.
___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 30 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyV-ptDES_IHPDu-yQL:-NyWEz9h6sNlUWqBKmHz:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/39d59e88c3591941a74f75429b4baf417e5dd44b/crates/blockifier/src/concurrency/flow_test.rs#L30)):*
> ```XML
>     // they both read from and write to the same storage entry, this couples the tx_index with the
>     // write set of the corresponding tx. However, to make the test more noisy, we determine the
>     // write-set dynamically.
> ```

Done.
___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 54 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyV2-oW2wF5tGe4_ZWk:-NyWF5qq43lfeIPGHaw6:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/39d59e88c3591941a74f75429b4baf417e5dd44b/crates/blockifier/src/concurrency/flow_test.rs#L54)):*
> ```XML
>                         .unwrap();
>                         // Simulate a write operation of a tx.
>                         let writes = state_maps_with_single_storage_entry(
> ```

Done.
___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 73 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyUwbA_AyqW06Jl3q_F:-NyWFb6y38G_3PPeR8qo:bs9b7nm) ([raw file](https://github.com/starkware-libs/blockifier/blob/39d59e88c3591941a74f75429b4baf417e5dd44b/crates/blockifier/src/concurrency/flow_test.rs#L73)):*

You can be more restrictive - the value must be there in validate phase.
</blockquote></details>

When `tx_index == 0` the `.get(&(contract_address, storage_key))` returns `None`. In this case, we need to read this one somehow.
I can change the comment to explain this specific case, WDYT?
___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 77 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyVYhX4AhGcM8--11sx:-NyWGcRoDKQa20sZIBQx:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/39d59e88c3591941a74f75429b4baf417e5dd44b/crates/blockifier/src/concurrency/flow_test.rs#L77)):*
> ```XML
>                         .unwrap();
>                         // If tx number tx_index wrote tx_written_value then it must be read
>                         // tx_written_value - 1.
> ```

Done.
___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 130 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyV2PakBCQe3gHBn0Bq:-NyWGoy0AXP28t1N5gF8:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/39d59e88c3591941a74f75429b4baf417e5dd44b/crates/blockifier/src/concurrency/flow_test.rs#L130)):*
> ```XML
> fn state_maps_with_single_storage_entry(
>     // tx_index: TxIndex,
>     contract_address: ContractAddress,
> ```

Done.

avi-starkware (commented):
> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)* status: all files reviewed, 6 unresolved discussions (waiting on @barak-b-starkware, @noaov1, and @Yoni-Starkware)

Yoni-Starkware (changes requested):
> Reviewed 1 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)* status: all files reviewed, 3 unresolved discussions (waiting on @barak-b-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 73 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyUwbA_AyqW06Jl3q_F:-NyWr5YhDq5aVRux4WKr:b-fqs6lb) ([raw file](https://github.com/starkware-libs/blockifier/blob/39d59e88c3591941a74f75429b4baf417e5dd44b/crates/blockifier/src/concurrency/flow_test.rs#L73)):*

When `tx_index == 0` the `.get(&(contract_address, storage_key))` returns `None`. In this case, we need to read this one somehow.
I can change the comment to explain this specific case, WDYT?
</blockquote></details>

Are you sure? It should return 1

avi-starkware (commented):
> Reviewed 4 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @barak-b-starkware and @noaov1)

avi-starkware (dismissed):
> Reviewed 1 of 5 files at r5.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)* status: all files reviewed, 1 unresolved discussion (waiting on @barak-b-starkware and @noaov1)

Yoni-Starkware (commented):
> Reviewed 2 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)* status: all files reviewed, 1 unresolved discussion (waiting on @barak-b-starkware and @noaov1)

barak-b-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)* status: 1 of 6 files reviewed, all discussions resolved (waiting on @avi-starkware, @noaov1, and @Yoni-Starkware)

___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 94 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-NyVZ-gb0MIEInsdSFlL:-NyZBE4T1zyaZ9OPX3Ub:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/39d59e88c3591941a74f75429b4baf417e5dd44b/crates/blockifier/src/concurrency/flow_test.rs#L94)):*

We changed the method `finish_validation` to `finish_abort` (to be used only if `aborted == true`)
</blockquote></details>

Done.

Yoni-Starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)* status: 1 of 6 files reviewed, 2 unresolved discussions (waiting on @avi-starkware, @barak-b-starkware, and @noaov1)

___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 64 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-O-g_VfRBWbpypYE4YuT:-O-g_VfRBWbpypYE4YuU:bwmu7oe) ([raw file](https://github.com/starkware-libs/blockifier/blob/beee66a428fb911a7b1960c7ab375dc3901f57c3/crates/blockifier/src/concurrency/flow_test.rs#L64)):*
> ```XML
>                         }
>                     }
>                 }
> ```

This is kind of duplicated and not relevant for this test, since nothing special is happening in the commit phase, but I can't think of an elegant halt condition



_Code quote:_
```XML
                if let Some(mut transaction_committer) = scheduler.try_enter_commit_phase() {
                    while let Some(tx_index) = transaction_committer.try_commit() {
                        let mut state_proxy = versioned_state.pin_version(tx_index);
                        let (reads, writes) =
                            get_reads_writes_for(Task::ValidationTask(tx_index), &versioned_state);
                        let reads_valid = state_proxy.validate_reads(&reads);
                        if !reads_valid {
                            state_proxy.delete_writes(&writes, &ContractClassMapping::default());
                            let (_, new_writes) = get_reads_writes_for(
                                Task::ExecutionTask(tx_index),
                                &versioned_state,
                            );
                            state_proxy.apply_writes(
                                &new_writes,
                                &ContractClassMapping::default(),
                                &HashMap::default(),
                            );
                            scheduler.finish_execution_during_commit(tx_index);
                        }
                    }
                }
```
___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 180 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-O-gZlaC3LxfAQVSB9Xd:-O-gZlaC3LxfAQVSB9Xe:b-lzvuke) ([raw file](https://github.com/starkware-libs/blockifier/blob/beee66a428fb911a7b1960c7ab375dc3901f57c3/crates/blockifier/src/concurrency/flow_test.rs#L180)):*
> ```XML
>     contract_address: ContractAddress,
>     storage_key: StorageKey,
>     value: u128,
> ```

Make the `contract_address` and the `storage_key` constant



_Suggestion:_
```XML
fn state_maps_with_single_value(value: u128)
```

barak-b-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)* status: 1 of 6 files reviewed, 1 unresolved discussion (waiting on @avi-starkware, @noaov1, and @Yoni-Starkware)

___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 64 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-O-g_VfRBWbpypYE4YuT:-O-jyyCRE9cyVm4rOP8D:b70t054) ([raw file](https://github.com/starkware-libs/blockifier/blob/beee66a428fb911a7b1960c7ab375dc3901f57c3/crates/blockifier/src/concurrency/flow_test.rs#L64)):*

This is kind of duplicated and not relevant for this test, since nothing special is happening in the commit phase, but I can't think of an elegant halt condition
</blockquote></details>

Why is it "kind of duplicated"?
I think it's relevant for this test because it aims to check **all** `pub` functions of the scheduler (and the transaction committer) **together** in one flow. And in this block, you have:
1. `scheduler.try_enter_commit_phase`
2. `transaction_committer.try_commit`
3. `scheduler.finish_execution_during_commit`

BTW, when running the test, you do get "re-executions" from the `if !reads_valid` block in the above quote.
___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 180 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-O-gZlaC3LxfAQVSB9Xd:-O-k3t2D-ohrVHfvu87f:b-kqzw2a) ([raw file](https://github.com/starkware-libs/blockifier/blob/beee66a428fb911a7b1960c7ab375dc3901f57c3/crates/blockifier/src/concurrency/flow_test.rs#L180)):*

Make the `contract_address` and the `storage_key` constant
</blockquote></details>

Done.
Notice that we can't use the fixture of `contrat_address` that way (I think it's a good suggestion to use the constant).
Maybe we can share the string from `test_utils`, instead of:
```
#[fixture]
pub fn contract_address() -> ContractAddress {
    contract_address!("0x18031991")
}
```
We could use the following:
```
const CONTRACT_ADDRESS: &str = "0x18031991";

#[fixture]
pub fn contract_address() -> ContractAddress {
    contract_address!(CONTRACT_ADDRESS)
}
```
LMK if you prefer this option

Yoni-Starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)* status: 1 of 6 files reviewed, all discussions resolved (waiting on @avi-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 64 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-O-g_VfRBWbpypYE4YuT:-O-k8d7b1OnkHeadsOkr:b-pr1t16) ([raw file](https://github.com/starkware-libs/blockifier/blob/beee66a428fb911a7b1960c7ab375dc3901f57c3/crates/blockifier/src/concurrency/flow_test.rs#L64)):*

Why is it "kind of duplicated"?
I think it's relevant for this test because it aims to check **all** `pub` functions of the scheduler (and the transaction committer) **together** in one flow. And in this block, you have:
1. `scheduler.try_enter_commit_phase`
2. `transaction_committer.try_commit`
3. `scheduler.finish_execution_during_commit`

BTW, when running the test, you do get "re-executions" from the `if !reads_valid` block in the above quote.
</blockquote></details>

In the sense that nothing new is happening in the commit phase (e.g., bouncer).
the following halt condition makes the commit logic unnecessary:
"all tasks were executed and validated"

The worker's `commit` also has annoying code duplication - it repeats validate and re-execute with small changes.

But I don't have a better solution, and as you said, you add coverage to the commit methods.

barak-b-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)* status: 1 of 6 files reviewed, all discussions resolved (waiting on @avi-starkware, @noaov1, and @Yoni-Starkware)

___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 64 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-O-g_VfRBWbpypYE4YuT:-O-kD1tc3IMZ2H-dbq3w:bdw611u) ([raw file](https://github.com/starkware-libs/blockifier/blob/beee66a428fb911a7b1960c7ab375dc3901f57c3/crates/blockifier/src/concurrency/flow_test.rs#L64)):*

In the sense that nothing new is happening in the commit phase (e.g., bouncer).
the following halt condition makes the commit logic unnecessary:
"all tasks were executed and validated"

The worker's `commit` also has annoying code duplication - it repeats validate and re-execute with small changes.

But I don't have a better solution, and as you said, you add coverage to the commit methods.
</blockquote></details>

Got it, so you want to merge it as is?

Yoni-Starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)* status: 1 of 6 files reviewed, all discussions resolved (waiting on @avi-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/flow_test.rs` line 64 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-O-g_VfRBWbpypYE4YuT:-O-kpjey6c5aw2DhqEUZ:b-b798d) ([raw file](https://github.com/starkware-libs/blockifier/blob/beee66a428fb911a7b1960c7ab375dc3901f57c3/crates/blockifier/src/concurrency/flow_test.rs#L64)):*

Alternatives:
1. Try to add the bouncer logic to this test (I think we shouldn't do that because even without the bouncer, you do check the re-execution flow of the commit mechanism)
2. I can think about how to eliminate the code duplication in the `worker`, and maybe the solution there can also be implemented in the test.
</blockquote></details>

We can merge it as is for now. I didn't approve since I didn't look at the entire code yet (if someone else approves, you can merge)

Yoni-Starkware (changes requested):
> Reviewed 3 of 5 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)* status: all files reviewed, 1 unresolved discussion (waiting on @barak-b-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/versioned_state.rs` line 248 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-O07eZtT5GrMp-UpU72z:-O07eZtU-uIIp9xh-fTU:b-n93714) ([raw file](https://github.com/starkware-libs/blockifier/blob/90e7e69b130cd2f5d2cfbcb7e0971855ed8c4238/crates/blockifier/src/concurrency/versioned_state.rs#L248)):*
> ```XML
>     pub fn state(&self) -> LockedVersionedState<'_, S> {
>         self.0.lock().expect("Failed to acquire state lock.")
>     }
> ```

Use use it only once. 
Remove and use into_inner_state instead (or `.lock().unwrap()` if something goes wrong)



_Code quote:_
```XML
    #[cfg(test)]
    pub fn state(&self) -> LockedVersionedState<'_, S> {
        self.0.lock().expect("Failed to acquire state lock.")
    }
```

Yoni-Starkware (commented):
One last comment

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-:-O07ewj08XkcmzlfZzK3:bvnowhg)* status: all files reviewed, 1 unresolved discussion (waiting on @barak-b-starkware and @noaov1)

barak-b-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830)* status: all files reviewed, 1 unresolved discussion (waiting on @noaov1 and @Yoni-Starkware)

___
*[`crates/blockifier/src/concurrency/versioned_state.rs` line 248 at r7](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-O07eZtT5GrMp-UpU72z:-O086WsvBBO2HPNzLKTi:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/90e7e69b130cd2f5d2cfbcb7e0971855ed8c4238/crates/blockifier/src/concurrency/versioned_state.rs#L248)):*

Use use it only once. 
Remove and use into_inner_state instead (or `.lock().unwrap()` if something goes wrong)
</blockquote></details>

Done.

Yoni-Starkware (approved):
:lgtm:

> Reviewed 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1830#-:-O088Hw37Rq952A7tjBi:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @noaov1)

PR #1863, opened 2024-05-06
feat(concurrency): add thread logic method

noaov1:


avi-starkware:
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 25 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxCu92fC3efxzt5fkyv:-NxCu92fC3efxzt5fkyw:undefined) ([raw file](https://github.com/starkware-libs/blockifier/blob/11f325db3720949e5b35cac5709d164cefcfe00d/crates/blockifier/src/concurrency/worker_logic.rs#L25)):*
> ```XML
>     loop {
>         task = match task {
>             Task::ExecutionTask(tx_index) => try_execute(
> ```





_Suggestion:_
```XML
            Task::ExecutionTask(tx_index) => execute(
```

avi-starkware:
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 35 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxCu5NL-8D32y2dYMv-:-NxCu5NL-8D32y2dYMv0:undefined) ([raw file](https://github.com/starkware-libs/blockifier/blob/11f325db3720949e5b35cac5709d164cefcfe00d/crates/blockifier/src/concurrency/worker_logic.rs#L35)):*
> ```XML
>             ),
>             Task::ValidationTask(tx_index) => {
>                 needs_reexecution(&scheduler, tx_index, &state, reads_writes)
> ```





_Suggestion:_
```XML
                validate(&scheduler, tx_index, &state, reads_writes)
```

barak-b-starkware:
*[`crates/blockifier/src/concurrency/versioned_state_proxy.rs` line 161 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxCyYaS-Gc9U8OdbgTR:-NxCyYaS-Gc9U8OdbgTS:b45h) ([raw file](https://github.com/starkware-libs/blockifier/blob/11f325db3720949e5b35cac5709d164cefcfe00d/crates/blockifier/src/concurrency/versioned_state_proxy.rs#L161)):*
> ```XML
>         ThreadSafeVersionedState(self.0.clone())
>     }
> }
> ```





_Suggestion:_
```XML
impl<S: StateReader> Clone for ThreadSafeVersionedState<S> {
    fn clone(&self) -> Self {
        ThreadSafeVersionedState(Arc::clone(&self.0))
    }
}
```

noaov1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1863)* status: all files reviewed, 3 unresolved discussions (waiting on @avi-starkware and @barak-b-starkware)

___
*[`crates/blockifier/src/concurrency/versioned_state_proxy.rs` line 161 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxCyYaS-Gc9U8OdbgTR:-NxD3JUw5Y1bZ8nZgBKF:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/11f325db3720949e5b35cac5709d164cefcfe00d/crates/blockifier/src/concurrency/versioned_state_proxy.rs#L161)):*
> ```XML
>         ThreadSafeVersionedState(self.0.clone())
>     }
> }
> ```

Done.
___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 25 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxCu92fC3efxzt5fkyv:-NxD3SXS6430FrNWWO4B:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/11f325db3720949e5b35cac5709d164cefcfe00d/crates/blockifier/src/concurrency/worker_logic.rs#L25)):*
> ```XML
>     loop {
>         task = match task {
>             Task::ExecutionTask(tx_index) => try_execute(
> ```

Done.
___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 35 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxCu5NL-8D32y2dYMv-:-NxD3Ssw25nf4yy5tY76:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/11f325db3720949e5b35cac5709d164cefcfe00d/crates/blockifier/src/concurrency/worker_logic.rs#L35)):*
> ```XML
>             ),
>             Task::ValidationTask(tx_index) => {
>                 needs_reexecution(&scheduler, tx_index, &state, reads_writes)
> ```

Done.

barak-b-starkware (commented):
> Reviewed 1 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1863)* status: 1 of 3 files reviewed, 10 unresolved discussions (waiting on @avi-starkware, @noaov1, and @Yoni-Starkware)

___
*[`crates/blockifier/src/concurrency.rs` line 6 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGUNV0DWsAWs92s3SQ:-NxGUNV17h5eiu0lzOSG:bfmu1qh) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency.rs#L6)):*
> ```XML
> pub mod versioned_state_proxy;
> pub mod versioned_storage;
> pub mod worker_logic;
> ```

I don't like the "worker" context here, it reminds me the workers from the Aerospike server or CI workers, but can live with that (:



_Suggestion:_
```XML
pub mod thread_logic;
```
___
*[`crates/blockifier/src/concurrency/versioned_state_proxy.rs` line 256 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGNAIg598ucf8K_CVL:-NxGNAIh5IR5RwqpQfZB:b-b9lv8f) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/versioned_state_proxy.rs#L256)):*
> ```XML
>     pub visited_pcs: HashMap<ClassHash, HashSet<usize>>,
>     pub result: TransactionExecutionResult<TransactionExecutionInfo>,
> }
> ```

Will the read-set (and maybe the write-set) not eventually be part of the `VersionedState`?



_Code quote:_
```XML
pub struct TxInputOutput {
    pub reads: StateMaps,
    pub writes: StateMaps,
    pub visited_pcs: HashMap<ClassHash, HashSet<usize>>,
    pub result: TransactionExecutionResult<TransactionExecutionInfo>,
}
```
___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 3 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGP06r6gz-ysua6JyV:-NxGP06r6gz-ysua6JyW:b7knvjr) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L3)):*
> ```XML
> use std::sync::Arc;
> 
> use cached::async_sync::Mutex;
> ```

Why this `Mutex` and not the `std::sync::Mutex`?



_Code quote:_
```XML
use cached::async_sync::Mutex;
```
___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 27 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGP7OY-slPV66A1upU:-NxGP7OZ4re95P_Sr5Cy:b7g2xq4) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L27)):*
> ```XML
>         task = match task {
>             Task::ExecutionTask(tx_index) => {
>                 execute(&scheduler, tx_index, &state, chunk, tx_input_output, block_context)
> ```

Can the `execute` `_chunk` argument be changed to `_tx: &Transaction` and do the following?
The same for `tx_input_output`, except for the `Mutex` type, which I'm not sure I know how to handle



_Suggestion:_
```XML
execute(&scheduler, tx_index, &state, &chunk[tx_index], &tx_input_output[tx_index].lock().<something that release it>, block_context)
```
___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 27 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGPTk03Cu-V1gRTDDZ:-NxGPTk16TmFdSUM9s26:b-sr3a6e) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L27)):*
> ```XML
>         task = match task {
>             Task::ExecutionTask(tx_index) => {
>                 execute(&scheduler, tx_index, &state, chunk, tx_input_output, block_context)
> ```

I think that the fact that `finish_execution()` doesn't return `Task` and the `finish_validation()` does is confusing and the right place to state this difference is in the upper (this) `loop`. WDYT? (including removing the return type of `execute`)



_Suggestion:_
```XML
execute(&scheduler, tx_index, &state, chunk, tx_input_output, block_context);
Task::NoTask
```
___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 33 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGRyk7C0C3BJ_Q4uzZ:-NxGRyk7C0C3BJ_Q4uz_:b-djebt1) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L33)):*
> ```XML
>             }
>             Task::NoTask => scheduler.next_task(),
>             Task::Done => todo!(),
> ```

Notice that it's only return after `scheduler.done()` call inside the `next_task()` method.
Or maybe we want a while loop that asks every time `!scheduler.done()` and in the `Task::Done` we return `()` (which is kind of optimization since you allow the thread to find out about it earlier (no need to call the last `next_task()`).



_Suggestion:_
```XML
Task::Done => break,
```
___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 45 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGSov1BC2Up_BOkfLE:-NxGSov1BC2Up_BOkfLF:undefined) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L45)):*
> ```XML
>     _tx_input_output: &[Mutex<TxInputOutput>],
>     _block_context: &BlockContext,
> ) -> Task {
> ```





_Suggestion:_
```XML
fn execute<S: StateReader>(
    _scheduler: &Scheduler,
    _tx_index: TxIndex,
    _state: &ThreadSafeVersionedState<S>,
    _tx: &Transaction,
    _tx_input_output: &TxInputOutput,
    _block_context: &BlockContext,
) {
```
___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 55 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGT45_C1e13-IdYoW4:-NxGT45_C1e13-IdYoW5:undefined) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L55)):*
> ```XML
>     _state: &ThreadSafeVersionedState<S>,
>     _tx_input_output: &[Mutex<TxInputOutput>],
> ) -> Task {
> ```





_Suggestion:_
```XML
fn validate<S: StateReader>(
    _scheduler: &Scheduler,
    _tx_index: TxIndex,
    _state: &ThreadSafeVersionedState<S>,
    _tx_input_output: &TxInputOutput,
) -> Task {
```

barak-b-starkware (changes requested):
> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1863)* status: all files reviewed, 10 unresolved discussions (waiting on @avi-starkware and @noaov1)

Yoni-Starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1863)* status: all files reviewed, 11 unresolved discussions (waiting on @avi-starkware, @barak-b-starkware, and @noaov1)

___
*[`crates/blockifier/src/concurrency.rs` line 6 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGUNV0DWsAWs92s3SQ:-NxGzYvo6PrH0dVSRUEY:b-9kp2ed) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency.rs#L6)):*

I don't like the "worker" context here, it reminds me the workers from the Aerospike server or CI workers, but can live with that (:
</blockquote></details>

We usually use `worker` to abstract its type (process/thread) 
___
*[`crates/blockifier/src/concurrency/versioned_state_proxy.rs` line 251 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxH-9vAEcT60xyzOfpo:-NxH-9vAEcT60xyzOfpp:bkyftoy) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/versioned_state_proxy.rs#L251)):*
> ```XML
> }
> 
> pub struct TxInputOutput {
> ```

I see only outputs here
(You can drop the `Task` if you want)



_Suggestion:_
```XML
pub struct ExecutionTaskOutput {
```
___
*[`crates/blockifier/src/concurrency/versioned_state_proxy.rs` line 256 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGNAIg598ucf8K_CVL:-NxH0BKc-hz89S8AxZUy:b-qsx9cn) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/versioned_state_proxy.rs#L256)):*

Will the read-set (and maybe the write-set) not eventually be part of the `VersionedState`?
</blockquote></details>

I think they should not be part of the VersionedState. It should be as thin as possible

Yoni-Starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1863)* status: all files reviewed, 12 unresolved discussions (waiting on @avi-starkware, @barak-b-starkware, and @noaov1)

___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 20 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxH0LsWDBfktqu4ImfE:-NxH0LsWDBfktqu4ImfF:b3kqx) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L20)):*
> ```XML
>     state: ThreadSafeVersionedState<S>,
>     chunk: &[Transaction],
>     tx_input_output: &[Mutex<TxInputOutput>],
> ```

.



_Suggestion:_
```XML
execution_outputs
```

avi-starkware:
*[`crates/blockifier/src/concurrency/versioned_state_proxy.rs` line 251 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxH-9vAEcT60xyzOfpo:-NxH0k7c6RO6JK9pBH5z:bwcx1ha) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/versioned_state_proxy.rs#L251)):*

I see only outputs here
(You can drop the `Task` if you want)
</blockquote></details>

reads are viewed as the inputs

noaov1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1863)* status: all files reviewed, 11 unresolved discussions (waiting on @avi-starkware, @barak-b-starkware, and @Yoni-Starkware)

___
*[`crates/blockifier/src/concurrency/versioned_state_proxy.rs` line 256 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGNAIg598ucf8K_CVL:-NxGyK4uE5YfR5T3N98i:b-l00yyj) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/versioned_state_proxy.rs#L256)):*

I think they should not be part of the VersionedState. It should be as thin as possible
</blockquote></details>

Remember that in the sate we have a "global lock". Maybe when we change the lock on the state we can consider doing that. 
___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 3 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGP06r6gz-ysua6JyV:-NxGyb4G7yfMhqxUuEhX:b-1bbggc) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L3)):*

Why this `Mutex` and not the `std::sync::Mutex`?
</blockquote></details>

Nice catch! For no good reason :)
___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 27 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGP7OY-slPV66A1upU:-NxGzbPM7PF1j7TT-BBw:b-mcev9c) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L27)):*

Can the `execute` `_chunk` argument be changed to `_tx: &Transaction` and do the following?
The same for `tx_input_output`, except for the `Mutex` type, which I'm not sure I know how to handle
</blockquote></details>

Why do you think it's better?
I think that for the `tx_input_output` it will be problematic since we don't want to take the lock for the entire method execution. 
___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 27 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGPTk03Cu-V1gRTDDZ:-NxH--qc0RfbgQWg0lBj:b-o5st0l) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L27)):*

I think that the fact that `finish_execution()` doesn't return `Task` and the `finish_validation()` does is confusing and the right place to state this difference is in the upper (this) `loop`. WDYT? (including removing the return type of `execute`)
</blockquote></details>

Maybe it's better to call directly the `next_task` method. WDYT?
___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 33 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGRyk7C0C3BJ_Q4uzZ:-NxH-az2Aso5_BKvbSKH:b-lgq90y) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L33)):*

Notice that it's only return after `scheduler.done()` call inside the `next_task()` method.
Or maybe we want a while loop that asks every time `!scheduler.done()` and in the `Task::Done` we return `()` (which is kind of optimization since you allow the thread to find out about it earlier (no need to call the last `next_task()`).
</blockquote></details>

Right. I thought that we might want to do something when we are done (commit the chunk changes into the block state), but maybe it's better to do it in the main loop. I'll break

Yoni-Starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1863)* status: all files reviewed, 11 unresolved discussions (waiting on @avi-starkware, @barak-b-starkware, and @noaov1)

___
*[`crates/blockifier/src/concurrency/versioned_state_proxy.rs` line 251 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxH-9vAEcT60xyzOfpo:-NxH1SqbA1_gQNRItjTN:bb76mu4) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/versioned_state_proxy.rs#L251)):*

reads are viewed as the inputs
</blockquote></details>

Why?

avi-starkware:
*[`crates/blockifier/src/concurrency/versioned_state_proxy.rs` line 256 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGNAIg598ucf8K_CVL:-NxH0uvL8a9wy5vir3ZN:bo88g71) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/versioned_state_proxy.rs#L256)):*

Remember that in the sate we have a "global lock". Maybe when we change the lock on the state we can consider doing that.
</blockquote></details>

The versioned state needs to manage everything that is happening behind the scenes when a read or a write happens. It doesn't make sense to put it in the executor. The executor simply talks to the state, doing writes and reads and validations. We might want to change the implementation of the versioned state and the executor should be oblivious to these implementation details.

avi-starkware:
*[`crates/blockifier/src/concurrency/versioned_state_proxy.rs` line 251 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxH-9vAEcT60xyzOfpo:-NxH1YAu527yTbF9jXem:byo8vly) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/versioned_state_proxy.rs#L251)):*

Why?
</blockquote></details>

Because they are external elements that affect execution. While writes are outputs because they remain as part of the state and they may affect the execution of other txs.

avi-starkware:
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 27 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGP7OY-slPV66A1upU:-NxH3LZw02JZh2ZKDaB_:b-mo42nj) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L27)):*

Why do you think it's better?
I think that for the `tx_input_output` it will be problematic since we don't want to take the lock for the entire method execution.
</blockquote></details>

I think it is cleaner to use the tx_index as the representative of the tx. We need the index anyway for when we communicate with the scheduler, and since no one needs write access to a tx, passing a ref to a single tx is the same as passing a ref to the entire chunk (same for tx_input_output).

avi-starkware:
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 27 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGPTk03Cu-V1gRTDDZ:-NxH4Luh4_noUWLOOeuC:bussmm7) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L27)):*

Maybe it's better to call directly the `next_task` method. WDYT?
</blockquote></details>

I think it is better to have the execute function output Task::NoTask (or maybe the `finish_execution()` method?). This way, the signature of the function will remain the same when we add dependencies.

noaov1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1863)* status: 2 of 4 files reviewed, 11 unresolved discussions (waiting on @avi-starkware, @barak-b-starkware, and @Yoni-Starkware)

___
*[`crates/blockifier/src/concurrency/versioned_state_proxy.rs` line 251 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxH-9vAEcT60xyzOfpo:-NxH4OTt2i7JFSeDVHsX:b-l2wdgj) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/versioned_state_proxy.rs#L251)):*

Because they are external elements that affect execution. While writes are outputs because they remain as part of the state and they may affect the execution of other txs.
</blockquote></details>

The reads are inputs to the validate method. 
___
*[`crates/blockifier/src/concurrency/versioned_state_proxy.rs` line 256 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGNAIg598ucf8K_CVL:-NxH4e4x52pR71I31iYm:b-s0dzej) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/versioned_state_proxy.rs#L256)):*
> The versioned state needs to manage everything that is happening behind the scenes when a read or a write happens.

Why?
___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 27 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGPTk03Cu-V1gRTDDZ:-NxH5xwg5_tooqADAKX4:bb6un1j) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L27)):*

I think it is better to have the execute function output Task::NoTask (or maybe the `finish_execution()` method?). This way, the signature of the function will remain the same when we add dependencies.
</blockquote></details>

Done

Yoni-Starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1863)* status: 2 of 4 files reviewed, 11 unresolved discussions (waiting on @avi-starkware, @barak-b-starkware, and @noaov1)

___
*[`crates/blockifier/src/concurrency/versioned_state_proxy.rs` line 251 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxH-9vAEcT60xyzOfpo:-NxHUyTl4UnMG6U7M5a2:b-eopbtq) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/versioned_state_proxy.rs#L251)):*

The reads are inputs to the validate method.
</blockquote></details>

They are all outputs of the execution task, and the execution task handler is the one that updates this array, right?

barak-b-starkware (changes requested):
> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1863)* status: all files reviewed, 10 unresolved discussions (waiting on @avi-starkware, @noaov1, and @Yoni-Starkware)

___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 27 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGPTk03Cu-V1gRTDDZ:-NxHfLfX5gFEONBTjtYo:b91q0iu) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L27)):*

Done
</blockquote></details>

This solution is hiding the difference even more, I think it’s important to notice that the execution can’t really return any task whereas the validation can. Making it implicit by returning Task::NoTask doesn’t make sense
___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 33 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGRyk7C0C3BJ_Q4uzZ:-NxHgGY58JeHu_zkRK25:b6rmjrc) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L33)):*

Right. I thought that we might want to do something when we are done (commit the chunk changes into the block state), but maybe it's better to do it in the main loop. I'll break
</blockquote></details>

So I think we can remove the pub from done

noaov1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1863)* status: all files reviewed, 6 unresolved discussions (waiting on @avi-starkware, @barak-b-starkware, and @Yoni-Starkware)

___
*[`crates/blockifier/src/concurrency/versioned_state_proxy.rs` line 251 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxH-9vAEcT60xyzOfpo:-NxHiCy150hniQgeg4VG:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/versioned_state_proxy.rs#L251)):*

They are all outputs of the execution task, and the execution task handler is the one that updates this array, right?
</blockquote></details>

Done.
___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 20 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxH0LsWDBfktqu4ImfE:-NxHiIbD3CBbABY8CCUE:b-896fix) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L20)):*

.
</blockquote></details>

Done.
___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 27 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGPTk03Cu-V1gRTDDZ:-NxHiKB_3mgpcs8KlyVn:b-oqmgdi) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L27)):*

This solution is hiding the difference even more, I think it’s important to notice that the execution can’t really return any task whereas the validation can. Making it implicit by returning Task::NoTask doesn’t make sense
</blockquote></details>

Okay, I see your point. I will challenge it (not sure I fully believe my words though): why do we have the no task variant for?  Don't we want the `execute` and `validate` to be consistent?
___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 33 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGRyk7C0C3BJ_Q4uzZ:-NxHjPdr4AVQUcTBUvfk:bb6un1j) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L33)):*

So I think we can remove the pub from done
</blockquote></details>

Done

Yoni-Starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1863)* status: 1 of 4 files reviewed, 4 unresolved discussions (waiting on @avi-starkware and @barak-b-starkware)

noaov1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1863)* status: 1 of 4 files reviewed, 5 unresolved discussions (waiting on @avi-starkware, @barak-b-starkware, and @Yoni-Starkware)

___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 32 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxIDPRp77yuyYr4eqsH:-NxIDPRp77yuyYr4eqsI:b-l6j0yq) ([raw file](https://github.com/starkware-libs/blockifier/blob/c5ed5bb7de945683504649215775e9a615e36339/crates/blockifier/src/concurrency/worker_logic.rs#L32)):*
> ```XML
> // TODO(Noa, 15/05/2024): Re-consider the necessity of the Arc (as opposed to a reference), given
> // concurrent code.
> pub fn run<S: StateReader>(execution_context: Arc<ConcurrentExecutionContext<S>>) {
> ```

I'm wondering if these methods should be of the struct `ConcurrentExecutionContext`.
Are the methods associated with that type? Are they encapsulate behavior directly related to that type? (I'm thinking no)
WDYT?



_Code quote:_
```XML
pub fn run<S: StateReader>(execution_context: Arc<ConcurrentExecutionContext<S>>) {
```

barak-b-starkware (changes requested):
> Reviewed 2 of 3 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1863)* status: all files reviewed, 6 unresolved discussions (waiting on @avi-starkware, @noaov1, and @Yoni-Starkware)

___
*[`crates/blockifier/src/concurrency/versioned_state_proxy.rs` line 256 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGNAIg598ucf8K_CVL:-NxIK9IP2RlGSDbwlBWo:bizcfx8) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/versioned_state_proxy.rs#L256)):*

> The versioned state needs to manage everything that is happening behind the scenes when a read or a write happens.

Why?
</blockquote></details>

I'm not sure I understand why you say that the reads shouldn't be part of the `VersionedState`
___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 27 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGP7OY-slPV66A1upU:-NxIKkum7ol-atYfogWj:b-dmwy8a) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L27)):*

I think it is cleaner to use the tx_index as the representative of the tx. We need the index anyway for when we communicate with the scheduler, and since no one needs write access to a tx, passing a ref to a single tx is the same as passing a ref to the entire chunk (same for tx_input_output).
</blockquote></details>

Relevant for passing all arguments explicitly (not by `execution_context`)
Does the `execute` method (bad name, BTW - overloading-wise) need access to the entire chunk? I don't understand why the entire chunk should be passed if the answer is no.
Same for `tx_input_output`, re the lock - IIUC we lock only one cell, no?
___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 27 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGPTk03Cu-V1gRTDDZ:-NxILfqC97mqbf5wy4_z:b-yrsn0m) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L27)):*

Okay, I see your point. I will challenge it (not sure I fully believe my words though): why do we have the no task variant for?  Don't we want the `execute` and `validate` to be consistent?
</blockquote></details>

I'm not sure why we have it, I remember that @avi-starkware suggested adding it.
Anyway, `next_task` here "is not fair" and IMO not so beautiful, even though it stresses the difference. Second, it's not consistent with the paper (weak argument but still). I prefer returning `Task::NoTask`.
Was it convincing enough?
___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 21 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxIMpV-9kgJAAMzxfQu:-NxIMpV-9kgJAAMzxfQv:b8k2sgl) ([raw file](https://github.com/starkware-libs/blockifier/blob/c5ed5bb7de945683504649215775e9a615e36339/crates/blockifier/src/concurrency/worker_logic.rs#L21)):*
> ```XML
>     pub execution_outputs: Box<[Mutex<ExecutionTaskOutput>]>,
>     pub block_context: BlockContext,
> }
> ```

Why did you move all of them into a struct?



_Code quote:_
```XML
pub struct ConcurrentExecutionContext<S: StateReader> {
    pub scheduler: Scheduler,
    pub state: ThreadSafeVersionedState<S>,
    pub chunk: Box<[Transaction]>,
    pub execution_outputs: Box<[Mutex<ExecutionTaskOutput>]>,
    pub block_context: BlockContext,
}
```

Yoni-Starkware (commented):
> Reviewed 1 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1863)* status: all files reviewed, 6 unresolved discussions (waiting on @avi-starkware, @barak-b-starkware, and @noaov1)

___
*[`crates/blockifier/src/concurrency/versioned_state_proxy.rs` line 256 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGNAIg598ucf8K_CVL:-NxIaJ3oApoBGWAfClrU:bdjvehp) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/versioned_state_proxy.rs#L256)):*

I'm not sure I understand why you say that the reads shouldn't be part of the `VersionedState`
</blockquote></details>

* It's not necessary. The API (get, apply_writes, undo, etc) does not assume the existence of these objects, and I think it's more elegant that it holds just the versioned storage cells.
* The writes (and maybe the reads as well?) are coupled with the results and visited pcs. Putting all of them in the versioned state makes less sense to me.
* In our case, it's better to save the lock needed for accessing it (as Noa said)
___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 21 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxIMpV-9kgJAAMzxfQu:-NxIaefoDBS2t-BezHE8:b-t5cejd) ([raw file](https://github.com/starkware-libs/blockifier/blob/c5ed5bb7de945683504649215775e9a615e36339/crates/blockifier/src/concurrency/worker_logic.rs#L21)):*

Why did you move all of them into a struct?
</blockquote></details>

It saves passing a log of arguments. Why not?

Yoni-Starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1863)* status: all files reviewed, 6 unresolved discussions (waiting on @avi-starkware, @barak-b-starkware, and @noaov1)

___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 32 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxIDPRp77yuyYr4eqsH:-NxIctU7ElvwtKn02suo:byhd3wz) ([raw file](https://github.com/starkware-libs/blockifier/blob/c5ed5bb7de945683504649215775e9a615e36339/crates/blockifier/src/concurrency/worker_logic.rs#L32)):*

I'm wondering if these methods should be of the struct `ConcurrentExecutionContext`.
Are the methods associated with that type? Are they encapsulate behavior directly related to that type? (I'm thinking no)
WDYT?
</blockquote></details>

Looks good right now, let's see in the future

avi-starkware:
*[`crates/blockifier/src/concurrency/versioned_state_proxy.rs` line 256 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGNAIg598ucf8K_CVL:-NxItspz9fdrrL9PlQid:bb71kmx) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/versioned_state_proxy.rs#L256)):*

* It's not necessary. The API (get, apply_writes, undo, etc) does not assume the existence of these objects, and I think it's more elegant that it holds just the versioned storage cells.
* The writes (and maybe the reads as well?) are coupled with the results and visited pcs. Putting all of them in the versioned state makes less sense to me.
* In our case, it's better to save the lock needed for accessing it (as Noa said)
</blockquote></details>

Okay

noaov1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1863)* status: all files reviewed, 4 unresolved discussions (waiting on @avi-starkware and @barak-b-starkware)

___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 27 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGP7OY-slPV66A1upU:-NxM9FrqA3ODpF0XKd_x:b-i8qmmz) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L27)):*

Relevant for passing all arguments explicitly (not by `execution_context`)
Does the `execute` method (bad name, BTW - overloading-wise) need access to the entire chunk? I don't understand why the entire chunk should be passed if the answer is no.
Same for `tx_input_output`, re the lock - IIUC we lock only one cell, no?
</blockquote></details>

The execute method does not need the entire chunk.
However, as Avi said, it gets the `tx_index` so it makes sense to use it. Plus, that way, you ensure the coupling between all the transaction relevant objects. 
Renaming can be done later. 
___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 27 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGPTk03Cu-V1gRTDDZ:-NxLwcAI9l0Dq32Ly7Zt:b-4tfbn6) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L27)):*

I'm not sure why we have it, I remember that @avi-starkware suggested adding it.
Anyway, `next_task` here "is not fair" and IMO not so beautiful, even though it stresses the difference. Second, it's not consistent with the paper (weak argument but still). I prefer returning `Task::NoTask`.
Was it convincing enough?
</blockquote></details>

We need it in case there is not task (see [here](https://github.com/starkware-libs/blockifier/blob/b540c603fb01a0d46b8246a5f860c40dc3134a80/crates/blockifier/src/concurrency/scheduler.rs#L75)). 
Why "not fair"? We know that the execution task will not return a new task, so it makes sense IMO to call the `next_task`.
If you ask me, if you want to return `Task::NoTask`, it makes more sense for `finish_execution` to return it.
Anyway, I don't think it matters that much. Changed.
___
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 21 at r6](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxIMpV-9kgJAAMzxfQu:-NxMBJBl4_LgsVjCRbBV:b4kj4xp) ([raw file](https://github.com/starkware-libs/blockifier/blob/c5ed5bb7de945683504649215775e9a615e36339/crates/blockifier/src/concurrency/worker_logic.rs#L21)):*

It saves passing a log of arguments. Why not?
</blockquote></details>

+ 1.
In addition, all those fields are considered the execution context of the thread so it makes sense to put them together.

barak-b-starkware (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-:-NxMR-FR86pN93VsC5aU:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @avi-starkware)

Yoni-Starkware (approved):
> Reviewed 1 of 3 files at r5, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1863)* status: all files reviewed, 2 unresolved discussions (waiting on @avi-starkware)

avi-starkware:
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 27 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGPTk03Cu-V1gRTDDZ:-NxJ2keh0P21JTYTsXyL:b-q8e1cr) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L27)):*

I'm not sure why we have it, I remember that @avi-starkware suggested adding it.
Anyway, `next_task` here "is not fair" and IMO not so beautiful, even though it stresses the difference. Second, it's not consistent with the paper (weak argument but still). I prefer returning `Task::NoTask`.
Was it convincing enough?
</blockquote></details>

`NoTask` was added to avoid having all tasks be optional and then needing to wrap everything with Some which is annoying.

avi-starkware:
*[`crates/blockifier/src/concurrency/worker_logic.rs` line 27 at r3](https://reviewable.io/reviews/starkware-libs/blockifier/1863#-NxGP7OY-slPV66A1upU:-NxIuSCA7SUj8RrvkQHw:bxwjwfg) ([raw file](https://github.com/starkware-libs/blockifier/blob/8c97b2b369b9260f6ad0e9009e2e1c9a939dfe24/crates/blockifier/src/concurrency/worker_logic.rs#L27)):*

Relevant for passing all arguments explicitly (not by `execution_context`)
Does the `execute` method (bad name, BTW - overloading-wise) need access to the entire chunk? I don't understand why the entire chunk should be passed if the answer is no.
Same for `tx_input_output`, re the lock - IIUC we lock only one cell, no?
</blockquote></details>

I don't think it matters that much. It is a matter of taste. The thread borrows the entire array anyway, so there is no harm in passing it through to the function that actually does something. Additionally, it makes things more consistent: throughout the code, we hold the entire chunk and access the tx we want via the tx_index. This may prevent a bug where we accidentally create a discrepancy somewhere - passing a transaction along with the wrong tx_index, which would lead to unexpected behavior.

Regarding the name `execute`: if you mean changing execute to "run" / "apply" as we discussed, we should do it across the board in one PR. If not, this is not overloading: when you get an execution task the operation you perform is "execute"; when you get a validation task the operation you perform is "validate". I think it makes the most sense.

PR #1889, opened 2024-05-08
feat(concurrency): add commit transaction functionality to the scheduler

avi-starkware:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1889)

### DIFF
--- a/crates/blockifier/src/concurrency/scheduler.rs
+++ b/crates/blockifier/src/concurrency/scheduler.rs
@@ -12,16 +12,24 @@ pub mod test;
 pub struct Scheduler {
     execution_index: AtomicUsize,
     validation_index: AtomicUsize,
-    /// Read twice upon checking the chunk completion. Used to detect if validation or execution
-    /// index decreased from their observed values after ensuring that the number of active tasks
-    /// is zero.
+    // The index of the next transaction to commit. `commit_index` is protected by `commit_lock`.
+    // There will be no concurrent access to `commit_index`, but the rust compiler cannot infer
+    // that this is the case, so we use a Mutex to allow mutable and concurrent access.
+    // TODO(Avi, 15/05/2024): Consider defining an ExplicitSyncWrapper for this.
+    commit_index: Mutex<usize>,
+    // Used to lock the commit index when committing transactions. This is necessary to make sure
+    // the process of committing transactions is sequential.
+    commit_lock: AtomicBool,
+    // Read twice upon checking the chunk completion. Used to detect if validation or execution
+    // index decreased from their observed values after ensuring that the number of active tasks
+    // is zero.
     decrease_counter: AtomicUsize,
     n_active_tasks: AtomicUsize,
     chunk_size: usize,
     // TODO(Avi, 15/05/2024): Consider using RwLock instead of Mutex.
     tx_statuses: Box<[Mutex<TransactionStatus>]>,
-    /// Updated by the `check_done` procedure, providing a cheap way for all threads to exit their
-    /// main loops.
+    // Updated by the `check_done` procedure, providing a cheap way for all threads to exit their
+    // main loops.
     done_marker: AtomicBool,
 }
 
@@ -30,6 +38,8 @@ impl Scheduler {
         Scheduler {
             execution_index: AtomicUsize::new(0),
             validation_index: AtomicUsize::new(chunk_size),
+            commit_index: Mutex::new(0),
+            commit_lock: AtomicBool::new(false),
             decrease_counter: AtomicUsize::new(0),
             n_active_tasks: AtomicUsize::new(0),
             chunk_size,
@@ -105,6 +115,49 @@ impl Scheduler {
         self.safe_decrement_n_active_tasks();
 
         Task::NoTask
+    }
+
+    /// Tries to commit the next transaction in the chunk. Returns the index of the transaction to
+    /// commit if successful, or None if the transaction is not yet executed.
+    /// Assumes that the caller has already acquired the commit lock.
+    pub fn try_commit(&self) -> Option<usize> {
+        let mut commit_index = self.commit_index.lock().expect(
+            "This lock should always succeed, since commit_lock must be successfully acquired \
+             before calling this method",
+        );
+        let mut status = self.lock_tx_status(*commit_index);
+        if *status == TransactionStatus::Executed {
+            *status = TransactionStatus::Committed;
+            *commit_index += 1;
+            if *commit_index == self.chunk_size {
+                self.done_marker.store(true, Ordering::Release);
+            }
+            return Some(*commit_index - 1);
+        }
+        None
+    }
+
+    /// This method is called after a transaction gets re-executed during a commit. It decreases the
+    /// validation index to ensure that higher transactions are validated. There is no need to set
+    /// the transaction status to Executed, as it is already set to Committed.
+    // TODO(Meshi, 01/06/2024): Add a call to this method after re-executing a transaction during
+    // commit.
+    pub fn finish_execution_during_commit(&self, tx_index: TxIndex) {
+        self.decrease_validation_index(tx_index + 1);
+    }
+
+    // TODO(Avi, 15/05/2024): Remove dead_code attribute after using the commit lock.
+    #[allow(dead_code)]
+    fn try_commit_lock(&self) -> bool {
+        self.commit_lock
+            .compare_exchange_weak(false, true, Ordering::Acquire, Ordering::Relaxed)
+            .is_ok()
+    }
+
+    // TODO(Avi, 15/05/2024): Remove dead_code attribute after using the commit lock.
+    #[allow(dead_code)]
+    fn unlock_commit_lock(&self) {
+        self.commit_lock.store(false, Ordering::Release);
     }
 
     /// Checks if all transactions have been executed and validated.
@@ -235,4 +288,5 @@ enum TransactionStatus {
     Executing,
     Executed,
     Aborting,
+    Committed,
 }


meship-starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1889)* status: all files reviewed, 1 unresolved discussion (waiting on @avi-starkware, @barak-b-starkware, and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 132 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-NxRcQFv6AIK-vdXLnMX:-NxRcQFv6AIK-vdXLnMY:bp89pk5) ([raw file](https://github.com/starkware-libs/blockifier/blob/139a0103faef1e6a0751b02ea326c5f67f6c66c9/crates/blockifier/src/concurrency/scheduler.rs#L132)):*
> ```XML
>              before calling this method",
>         );
>         // FIXME: Do we want the thread to wait for the status lock to be available?
> ```

Why not? Can a status be locked for a long time?

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 132 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-NxRcQFv6AIK-vdXLnMX:-NxReTNA8pVEwk7WYQyw:b-wztwk5) ([raw file](https://github.com/starkware-libs/blockifier/blob/139a0103faef1e6a0751b02ea326c5f67f6c66c9/crates/blockifier/src/concurrency/scheduler.rs#L132)):*
<details><summary><i>Previously, meship-starkware (Meshi Peled) wrote…</i></summary><blockquote>

Why not? Can a status be locked for a long time?
</blockquote></details>

We might want to have the thread move on if the status lock is taken, to not have it stuck doing nothing.

meship-starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1889)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @barak-b-starkware and @noaov1)

noaov1 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1889)* status: all files reviewed, 4 unresolved discussions (waiting on @avi-starkware and @barak-b-starkware)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 17 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-Nxug2D133eiEWKz2taK:-Nxug2D133eiEWKz2taL:b168y5s) ([raw file](https://github.com/starkware-libs/blockifier/blob/c62e022c86d907510571e2e04f6d5505da684c9a/crates/blockifier/src/concurrency/scheduler.rs#L17)):*
> ```XML
>     // The index of the next transaction to commit. `commit_index` is protected by `commit_lock`.
>     // There will be no concurrent access to `commit_index`, but the rust compiler cannot infer
>     // that this is the case, so we use a Mutex to protect it.
> ```

We use the mutex for the mutability, no?



_Code quote:_
```XML
    // There will be no concurrent access to `commit_index`, but the rust compiler cannot infer
    // that this is the case, so we use a Mutex to protect it.
```
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 18 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-Nxug7T75PYHNCQCbN7m:-Nxug7T75PYHNCQCbN7n:b-agweh3) ([raw file](https://github.com/starkware-libs/blockifier/blob/c62e022c86d907510571e2e04f6d5505da684c9a/crates/blockifier/src/concurrency/scheduler.rs#L18)):*
> ```XML
>     // There will be no concurrent access to `commit_index`, but the rust compiler cannot infer
>     // that this is the case, so we use a Mutex to protect it.
>     // TODO(Avi, 15/05/2024): Consider defining an ExplicitSyncWrapper for this.
> ```

Please add it to Monday. Can you please explain WDYM by `ExplicitSyncWrapper`?



_Code quote:_
```XML
    // TODO(Avi, 15/05/2024): Consider defining an ExplicitSyncWrapper for this.
```
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 122 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-NxugxgwCmth0xAXJ0EA:-NxugxgwCmth0xAXJ0EB:b-8yq442) ([raw file](https://github.com/starkware-libs/blockifier/blob/c62e022c86d907510571e2e04f6d5505da684c9a/crates/blockifier/src/concurrency/scheduler.rs#L122)):*
> ```XML
>     /// Tries to commit the next transaction in the chunk. Returns the index of the transaction to
>     /// commit if successful, or None if the transaction is not yet executed.
>     /// Assumes that the caller has already acquired the commit lock.
> ```

Why not let it be a part of this method?



_Code quote:_
```XML
    /// Assumes that the caller has already acquired the commit lock.
```
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 134 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-NxuhLjCB0zAOJhrSk5P:-NxuhLjCB0zAOJhrSk5Q:b-h8i9vv) ([raw file](https://github.com/starkware-libs/blockifier/blob/c62e022c86d907510571e2e04f6d5505da684c9a/crates/blockifier/src/concurrency/scheduler.rs#L134)):*
> ```XML
>             *commit_index += 1;
>             if *commit_index == self.chunk_size {
>                 self.done_marker.store(true, Ordering::Release);
> ```

Too soon, no? We want to wait for the last commit to be finished



_Code quote:_
```XML
                self.done_marker.store(true, Ordering::Release);
```

avi-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1889)* status: all files reviewed, 4 unresolved discussions (waiting on @barak-b-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 17 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-Nxug2D133eiEWKz2taK:-NxuyADb3XzaHwGdLJvB:b-dzw8ap) ([raw file](https://github.com/starkware-libs/blockifier/blob/c62e022c86d907510571e2e04f6d5505da684c9a/crates/blockifier/src/concurrency/scheduler.rs#L17)):*

We use the mutex for the mutability, no?
</blockquote></details>

Theoretically, I would have wanted to use `usize` here without a mutex, but rust doesn't like it because it is not thread-safe - rust doesn't know that `commit_lock` protects `commit_index` because it is not enforced by the compiler.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 18 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-Nxug7T75PYHNCQCbN7m:-NxuyqUtCLlcbrcXRzVK:b-qbqenz) ([raw file](https://github.com/starkware-libs/blockifier/blob/c62e022c86d907510571e2e04f6d5505da684c9a/crates/blockifier/src/concurrency/scheduler.rs#L18)):*

Please add it to Monday. Can you please explain WDYM by `ExplicitSyncWrapper`?
</blockquote></details>

This is what aptos did. It means wrapping the usize value with an `Unsafe` wrapper so that the rust compiler does not enforce thread safety on this value.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 122 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-NxugxgwCmth0xAXJ0EA:-NxuznP3AyzOdcwWfsNM:br2l165) ([raw file](https://github.com/starkware-libs/blockifier/blob/c62e022c86d907510571e2e04f6d5505da684c9a/crates/blockifier/src/concurrency/scheduler.rs#L122)):*

Why not let it be a part of this method?
</blockquote></details>

Because we want the thread that took the `commit_lock` to keep trying to commit the next tx until it no longer can. If we take the lock as part of try_commit, it means we have to release it after each committed tx, which is unnecessary overhead.
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 134 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-NxuhLjCB0zAOJhrSk5P:-Nxv-hlB6BFwv12s42GA:b-ynir0n) ([raw file](https://github.com/starkware-libs/blockifier/blob/c62e022c86d907510571e2e04f6d5505da684c9a/crates/blockifier/src/concurrency/scheduler.rs#L134)):*

Too soon, no? We want to wait for the last commit to be finished
</blockquote></details>

No. Once a transaction gains the status committed, it can no longer switch status, so at this point, all transactions have been committed, and there is nothing more for the scheduler to do. The executor might end up not committing the final tx, but if that happens there is no reason to consult the scheduler any more - to get the index of the first transaction to be included in the next block, simply use the index of the transaction that the bouncer blocked.

noaov1 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1889)* status: all files reviewed, 3 unresolved discussions (waiting on @avi-starkware and @barak-b-starkware)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 17 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-Nxug2D133eiEWKz2taK:-NxwNO8IBca10BF2-rjZ:bnsbol8) ([raw file](https://github.com/starkware-libs/blockifier/blob/c62e022c86d907510571e2e04f6d5505da684c9a/crates/blockifier/src/concurrency/scheduler.rs#L17)):*

Theoretically, I would have wanted to use `usize` here without a mutex, but rust doesn't like it because it is not thread-safe - rust doesn't know that `commit_lock` protects `commit_index` because it is not enforced by the compiler.
</blockquote></details>

Using `&mut` will be okay with the compiler? A mutex is a good solution for the mutability need. 
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 128 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-NxwOvQCE_bYFrPurQJs:-NxwOvQCE_bYFrPurQJt:b-t44zby) ([raw file](https://github.com/starkware-libs/blockifier/blob/c62e022c86d907510571e2e04f6d5505da684c9a/crates/blockifier/src/concurrency/scheduler.rs#L128)):*
> ```XML
>              before calling this method",
>         );
>         // FIXME: Do we want the thread to wait for the status lock to be available?
> ```

I think that no. 



_Code quote:_
```XML
        // FIXME: Do we want the thread to wait for the status lock to be available?
```
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 146 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-NxwOXykClH2hVLw7zi-:-NxwOXyl6K6PeSoZMnxs:b-7zi8tj) ([raw file](https://github.com/starkware-libs/blockifier/blob/c62e022c86d907510571e2e04f6d5505da684c9a/crates/blockifier/src/concurrency/scheduler.rs#L146)):*
> ```XML
>     pub fn finish_execution_during_commit(&self, tx_index: TxIndex) {
>         // We skipped decreasing the validation index when the transaction was executed, so we do it
>         // now.
> ```

WDYM?



_Code quote:_
```XML
        // We skipped decreasing the validation index when the transaction was executed, so we do it
        // now.
```

noaov1 (dismissed):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1889)* status: all files reviewed, 3 unresolved discussions (waiting on @avi-starkware and @barak-b-starkware)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 128 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-NxwOvQCE_bYFrPurQJs:-NxwWnSbAqU1aeWZVYgR:b-zidg37) ([raw file](https://github.com/starkware-libs/blockifier/blob/c62e022c86d907510571e2e04f6d5505da684c9a/crates/blockifier/src/concurrency/scheduler.rs#L128)):*

I think that no.
</blockquote></details>

I did not understand the fixme: yes, we want the thread to wait for the lock status.

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 146 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-NxwOXykClH2hVLw7zi-:-Nxwa1gH4hYvDXCicneB:b3r53u2) ([raw file](https://github.com/starkware-libs/blockifier/blob/c62e022c86d907510571e2e04f6d5505da684c9a/crates/blockifier/src/concurrency/scheduler.rs#L146)):*

WDYM?
</blockquote></details>

This method is intended to be used after the `try_commit_transaction` method of the `WorkersExecutor` re-executed the transaction. This re-execution does not call the finish_execution method and thus does not decrease the validation index. Therefore, we must now decrease the validation index.

avi-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1889)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @barak-b-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 17 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-Nxug2D133eiEWKz2taK:-NxwcDWP3oWfD4mVXKdb:b398ihz) ([raw file](https://github.com/starkware-libs/blockifier/blob/c62e022c86d907510571e2e04f6d5505da684c9a/crates/blockifier/src/concurrency/scheduler.rs#L17)):*

Using `&mut` will be okay with the compiler? A mutex is a good solution for the mutability need.
</blockquote></details>

We need a lifetime specifier to try this solution. Is it worth trying?
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 128 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-NxwOvQCE_bYFrPurQJs:-NxwapCl13Ui7xrR1rHm:b-wd4adw) ([raw file](https://github.com/starkware-libs/blockifier/blob/c62e022c86d907510571e2e04f6d5505da684c9a/crates/blockifier/src/concurrency/scheduler.rs#L128)):*

I did not understand the fixme: yes, we want the thread to wait for the lock status.
</blockquote></details>

Okay. I removed the `FIXME`.

meship-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1889)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @avi-starkware, @barak-b-starkware, and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 146 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-NxwOXykClH2hVLw7zi-:-Ny-HTCtA0mCdP0V_Yrl:b-c62edl) ([raw file](https://github.com/starkware-libs/blockifier/blob/c62e022c86d907510571e2e04f6d5505da684c9a/crates/blockifier/src/concurrency/scheduler.rs#L146)):*

This method is intended to be used after the `try_commit_transaction` method of the `WorkersExecutor` re-executed the transaction. This re-execution does not call the finish_execution method and thus does not decrease the validation index. Therefore, we must now decrease the validation index.
</blockquote></details>

We only want to run it if there is a re-execution, right? Right now, there is no way to tell outside of try_commit_transaction if there was a re-execution. Should I add it as a TODO in the commit PR, or should the return value indicate whether there was a re-execution or not?

noaov1 (dismissed):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1889)* status: all files reviewed, 1 unresolved discussion (waiting on @avi-starkware, @barak-b-starkware, and @meship-starkware)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 17 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-Nxug2D133eiEWKz2taK:-Ny-Fdk7297A8Kb2quBo:b32w12) ([raw file](https://github.com/starkware-libs/blockifier/blob/c62e022c86d907510571e2e04f6d5505da684c9a/crates/blockifier/src/concurrency/scheduler.rs#L17)):*

We need a lifetime specifier to try this solution. Is it worth trying?
</blockquote></details>

No
___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 146 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-NxwOXykClH2hVLw7zi-:-Ny-FxHY95C55G5Hymkk:bcnd9we) ([raw file](https://github.com/starkware-libs/blockifier/blob/c62e022c86d907510571e2e04f6d5505da684c9a/crates/blockifier/src/concurrency/scheduler.rs#L146)):*
<details><summary><i>Previously, meship-starkware (Meshi Peled) wrote…</i></summary><blockquote>

We only want to run it if there is a re-execution, right? Right now, there is no way to tell outside of try_commit_transaction if there was a re-execution. Should I add it as a TODO in the commit PR, or should the return value indicate whether there was a re-execution or not?
</blockquote></details>

I think the doc is a bit confusing.
The `try_commit` method (of the executor) will run it, no?

meship-starkware (approved):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1889)* status: all files reviewed, 1 unresolved discussion (waiting on @avi-starkware and @barak-b-starkware)

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 146 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-NxwOXykClH2hVLw7zi-:-Ny-YcUoBx699DRC4tXz:b20ary4) ([raw file](https://github.com/starkware-libs/blockifier/blob/c62e022c86d907510571e2e04f6d5505da684c9a/crates/blockifier/src/concurrency/scheduler.rs#L146)):*
@meship-starkware We already added a TODO to add `finish_execution_during_commit` in your PR.
@noaov1 The `try_commit_transaction` method will run it in the flow where the validation failed during commit, and the transaction had to be re-executed.
> I think the doc is a bit confusing.
Do you mean the docstring of the method? Or this comment? I think we can delete the comment because decreasing the validation index is already mentioned in the docstring.

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 146 at r2](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-NxwOXykClH2hVLw7zi-:-Ny-_bghE_QI0svpy3Ha:b-qfzpxi) ([raw file](https://github.com/starkware-libs/blockifier/blob/c62e022c86d907510571e2e04f6d5505da684c9a/crates/blockifier/src/concurrency/scheduler.rs#L146)):*

@meship-starkware We already added a TODO to add `finish_execution_during_commit` in your PR.
@noaov1 The `try_commit_transaction` method will run it in the flow where the validation failed during commit, and the transaction had to be re-executed.
> I think the doc is a bit confusing.
Do you mean the docstring of the method? Or this comment? I think we can delete the comment because decreasing the validation index is already mentioned in the docstring.
</blockquote></details>

I removed the comment.

noaov1 (commented):
> Reviewed 1 of 1 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1889)* status: all files reviewed, 1 unresolved discussion (waiting on @avi-starkware and @barak-b-starkware)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 153 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-Ny06qD86IEvRtjDW-j7:-Ny06qD86IEvRtjDW-j8:b-fkcqfp) ([raw file](https://github.com/starkware-libs/blockifier/blob/725d6a8e24cfa8ff472d8df593a9859372ecde53/crates/blockifier/src/concurrency/scheduler.rs#L153)):*
> ```XML
>     fn try_commit_lock(&self) -> bool {
>         self.commit_lock
>             .compare_exchange_weak(false, true, Ordering::Acquire, Ordering::Relaxed)
> ```

Can you please explain your ordering choices?
Why, for instance, not do as follow: (as we do in other atomic variables, using`fetch_add` and `load` accordingly)



_Suggestion:_
```XML
            .compare_exchange_weak(false, true, Ordering::SeqCst, Ordering::Acquire)
```

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 153 at r5](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-Ny06qD86IEvRtjDW-j7:-Ny09Sf2-eMmmOybkggo:b-8oyjjt) ([raw file](https://github.com/starkware-libs/blockifier/blob/725d6a8e24cfa8ff472d8df593a9859372ecde53/crates/blockifier/src/concurrency/scheduler.rs#L153)):*

Can you please explain your ordering choices?
Why, for instance, not do as follow: (as we do in other atomic variables, using`fetch_add` and `load` accordingly)
</blockquote></details>

The first ordering parameter is the success ordering. That is, this is the ordering used in case the lock is was `false` and we changed it to `true`.
Setting it to `Acquire` means that the "load" part of this operation will have ordering `Acquire` and the "store" part will have ordering `Relaxed`.
The load ordering ensures that all data changes made by the thread that last stored a value in `commit_lock` will be ordered before the storing of that value, i.e., they will be visible to the current thread. Since this previously stored value is always going to be `false`, the last thread that stored a value in `commit_lock` will be the thread that released the commit lock last. Hence, we want the ordering to be at least `Acquire` for the load operation to be sure that the current thread continues the work of the previous thread that held the lock in a consistent manner. (Note that the ordering of the "store" in `unlock_commit_lock` is `Release` to complement this reasoning.)

The "store" part of the success case can indeed be `Relaxed`. In the success case, we always store `true` taking the lock. I claim we only want this operation to be atomic, and we don't care about synchronization with other threads. Since the operation is atomic, only one thread can take the lock. Any thread that tried to take the lock while it was already taken (i.e., the value was `true`) should basically ignore the existence of the thread that took the lock and behave as if it never tried to take the lock in the first place. No special ordering should be imposed between the thread that succeeded in taking the lock and the thread that failed.

The second ordering parameter is the failure ordering. That is, the ordering used for the load operation in case the lock was "true". For the same reason as in the "store" of the success case, the failure ordering should also be `Relaxed` (considering the viewpoint of the "load" of the thread that didn't take the lock instead of the "store" of the thread that did take the lock).

noaov1 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1889#-:-Ny0f_D5BMLL-iZhmI7C:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @barak-b-starkware)

PR #1937, opened 2024-05-30
Add deserialization for `ContractClass`

tcoratger:
## Description

This pull request adds deserialization support for the `ContractClass` struct. The changes include implementing the `Deserialize` trait for `ContractClassV1` and updating the unit tests to verify the correct deserialization of both `ContractClassV0` and `ContractClassV1`.

**Implementation of `Deserialize` for `ContractClassV1`:**
   - The `Deserialize` trait has been implemented for `ContractClassV1` to support deserialization from JSON.
   - The deserialization process involves converting the input into a JSON value, then into a JSON string, and finally using the `try_from_json_string` method to deserialize into a `ContractClassV1` object.


crates/blockifier/src/execution/contract_class.rs
```diff
@@ -166,6 +166,24 @@ impl Deref for ContractClassV1 {
...
+
+        // Convert into a JSON string
+        let json_string = serde_json::to_string(&json_value)
+            .map_err(|err| DeserializationError::custom(err.to_string()))?;
+
+        // Use try_from_json_string to deserialize into ContractClassV1
+        ContractClassV1::try_from_json_string(&json_string)
+            .map_err(|err| DeserializationError::custom(err.to_string()))
+    }
+}
```
tdelabro, line 185: The true issue here is that cairo_vm::program doesn't implement serde.
So we have to go through all this. It's inefficient.

I opened an issue on the cairo-vm: https://github.com/lambdaclass/cairo-vm/issues/1660

I think you should push this one first, and then update in the blockifier once it is merged

tdelabro (changes requested):

crates/blockifier/src/execution/contract_class.rs
Eikix: This issue looks https://github.com/lambdaclass/cairo-vm/issues/1660 like it'll take a long time to be implemented and merged, how can we not be blocked by it?

PR #1978, opened 2024-06-12
fix(concurrency): allow threads in pyo3 (release the GIL)

noaov1:


Yoni-Starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1978)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @avi-starkware, @barak-b-starkware, and @noaov1)

___
*[`crates/native_blockifier/src/py_block_executor.rs` line 232 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1978#-O-FDksN1EGp6tu6zRc9:-O-FDksO5UOaCRVgMQlz:b-ollw11) ([raw file](https://github.com/starkware-libs/blockifier/blob/076d771fba9bb281c9906b5ea7b1d44c4dd2bd0e/crates/native_blockifier/src/py_block_executor.rs#L232)):*
> ```XML
>         // Run.
>         let results =
>             Python::with_gil(|py| py.allow_threads(|| self.tx_executor().execute_txs(&txs)));
> ```

Something is weird here. You take the GIL just to suspend it.
According to the documentation, it `Temporarily releases the GIL, thus allowing other Python threads to run.`

So sounds like there should not be a problem while the GIL is not taken (which should be our case)

Non-blocking since it works



_Code quote:_
```XML
            Python::with_gil(|py| py.allow_threads(|| self.tx_executor().execute_txs(&txs)))
```

Yoni-Starkware (approved):
:lgtm:

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1978#-:-O-FIWWH2ghLPVpbRM_l:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @avi-starkware, @barak-b-starkware, and @noaov1)

avi-starkware:
*[`crates/native_blockifier/src/py_block_executor.rs` line 232 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1978#-O-FDksN1EGp6tu6zRc9:-O-FOqQI5IFSOf7XrLTx:b9719r2) ([raw file](https://github.com/starkware-libs/blockifier/blob/076d771fba9bb281c9906b5ea7b1d44c4dd2bd0e/crates/native_blockifier/src/py_block_executor.rs#L232)):*
The problem is that the object `txs_with_class_infos` is of type `&PyAny`. See [here](https://pyo3.rs/main/doc/pyo3/marker/struct.python#method.allow_threads):

> -   Its lifetime represents the scope of holding the GIL which can be used to create Rust references that are bound to it, such as `&`[`PyAny`](https://pyo3.rs/main/doc/pyo3/types/struct.pyany "struct pyo3::types::PyAny").

Since the method `execute_txs` gets a reference to a Python object, we have to take the GIL before we can spawn threads.

However, as far as I understand, any access to a `&PyAny` requires taking the GIL again. @noaov1 are you sure this doesn't mean our threads get blocked by the GIL each time `txs` is read from?

Yoni-Starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1978)* status: all files reviewed, 1 unresolved discussion (waiting on @avi-starkware, @barak-b-starkware, and @noaov1)

___
*[`crates/native_blockifier/src/py_block_executor.rs` line 232 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1978#-O-FDksN1EGp6tu6zRc9:-O-FY0St6iuvbvCJJFx9:b4qlswk) ([raw file](https://github.com/starkware-libs/blockifier/blob/076d771fba9bb281c9906b5ea7b1d44c4dd2bd0e/crates/native_blockifier/src/py_block_executor.rs#L232)):*

The problem is that the object `txs_with_class_infos` is of type `&PyAny`. See [here](https://pyo3.rs/main/doc/pyo3/marker/struct.python#method.allow_threads):

> -   Its lifetime represents the scope of holding the GIL which can be used to create Rust references that are bound to it, such as `&`[`PyAny`](https://pyo3.rs/main/doc/pyo3/types/struct.pyany "struct pyo3::types::PyAny").

Since the method `execute_txs` gets a reference to a Python object, we have to take the GIL before we can spawn threads.

However, as far as I understand, any access to a `&PyAny` requires taking the GIL again. @noaov1 are you sure this doesn't mean our threads get blocked by the GIL each time `txs` is read from?
</blockquote></details>

Don't we consume it before calling the executor?

avi-starkware:
*[`crates/native_blockifier/src/py_block_executor.rs` line 232 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1978#-O-FDksN1EGp6tu6zRc9:-O-F_M-Z768NrLjV-bJ3:b-tulntm) ([raw file](https://github.com/starkware-libs/blockifier/blob/076d771fba9bb281c9906b5ea7b1d44c4dd2bd0e/crates/native_blockifier/src/py_block_executor.rs#L232)):*

Don't we consume it before calling the executor?
</blockquote></details>

You are right. `txs` is a purely rust variable. So we just need the `with_gil` to gain access to the `Python` token (`py`) from which we can run the method `allow_threads`.

Moreover, `with_gil` would not have been needed at all if the method `execute_txs` did not have pythonic objects as arguments.

avi-starkware (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1978#-:-O-FcqMW88K6D_gEkDFg:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @barak-b-starkware and @noaov1)

noaov1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/blockifier/1978)* status: all files reviewed, 1 unresolved discussion (waiting on @avi-starkware, @barak-b-starkware, and @Yoni-Starkware)

___
*[`crates/native_blockifier/src/py_block_executor.rs` line 232 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1978#-O-FDksN1EGp6tu6zRc9:-O-FpOeZ4w7ET5JSYNzi:b-3monhn) ([raw file](https://github.com/starkware-libs/blockifier/blob/076d771fba9bb281c9906b5ea7b1d44c4dd2bd0e/crates/native_blockifier/src/py_block_executor.rs#L232)):*

You are right. `txs` is a purely rust variable. So we just need the `with_gil` to gain access to the `Python` token (`py`) from which we can run the method `allow_threads`.

Moreover, `with_gil` would not have been needed at all if the method `execute_txs` did not have pythonic objects as arguments.
</blockquote></details>

You are right. I'm not sure why `allow_threads` is needed for running threads in rust. The [docs](https://pyo3.rs/v0.19.1/parallelism) suggest that rayon threads do not need to use `allow_threads`, maybe the std ones do?

avi-starkware:
*[`crates/native_blockifier/src/py_block_executor.rs` line 232 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1978#-O-FDksN1EGp6tu6zRc9:-O-FrNVj512Jw8O4pfDX:bejhvfw) ([raw file](https://github.com/starkware-libs/blockifier/blob/076d771fba9bb281c9906b5ea7b1d44c4dd2bd0e/crates/native_blockifier/src/py_block_executor.rs#L232)):*

You are right. I'm not sure why `allow_threads` is needed for running threads in rust. The [docs](https://pyo3.rs/v0.19.1/parallelism) suggest that rayon threads do not need to use `allow_threads`, maybe the std ones do?
</blockquote></details>

The docs say that "pure rust functions" can use threads without worrying about the GIL, but functions that accept Python objects as arguments need to take the GIL before they can spawn a thread.

noaov1:
I agree. I added the `allow_threads` code to a function that accepts Python objects but does *not* spawn any threads. It calls a rust method that spawns threads.
BTW- it does not work using rayon threads as well (got timeout).

noaov1:
*[`crates/native_blockifier/src/py_block_executor.rs` line 232 at r1](https://reviewable.io/reviews/starkware-libs/blockifier/1978#-O-FDksN1EGp6tu6zRc9:-O-FxBWv5XLR9INjSswp:b-rlre6m) ([raw file](https://github.com/starkware-libs/blockifier/blob/076d771fba9bb281c9906b5ea7b1d44c4dd2bd0e/crates/native_blockifier/src/py_block_executor.rs#L232)):*

The docs say that "pure rust functions" can use threads without worrying about the GIL, but functions that accept Python objects as arguments need to take the GIL before they can spawn a thread.
</blockquote></details>

I agree. I added the `allow_threads` code to a function that accepts Python objects but does *not* spawn any threads. It calls a rust method that spawns threads.
BTW- it does not work using rayon threads as well (got timeout).
Anyway it works :)

